PR #2, opened 2017-06-20
Feature/initial project setup

birom:


src/main/webapp/app/layouts/navbar/navbar.scss
```diff
@@ -68,3 +68,12 @@ Logo styles
         }
     }
 }
+
+/*=================================================================================
+Style guide customization
+=================================================================================== */
+.megamenu-menu {
+    .dropdown-item.active {
+        color: #ffffff;
```
michael-wirth: is this there a bootstrap variable we could use instead of hard coded colors?

we should fix the stylesheet project to use the `~` instead of the bridle relativ pahts `../../../` then we can import the alv-ch.scss instead and also make use of the variables

PR #7, opened 2017-07-13
Feature/occupation input component

birom:
Auto complete input compment for occupations.
input-screenshot-01


package.json
```diff
@@ -7,32 +7,32 @@
...
+    "@ng-bootstrap/ng-bootstrap": "^1.0.0-alpha.28",
+    "core-js": "^2.4.1",
     "jquery": "3.2.1",
     "ng-jhipster": "0.2.3",
     "ng2-webstorage": "1.8.0",
     "ngx-cookie": "1.0.0",
     "ngx-infinite-scroll": "0.4.1",
     "reflect-metadata": "0.1.10",
-    "rxjs": "5.4.0",
+    "rxjs": "^5.1.0",
```
michael-wirth: is version downgrading correct?

package.json
```diff
@@ -86,10 +86,10 @@
     "tslint": "4.5.1",
     "tslint-eslint-rules": "^4.1.1",
     "tslint-loader": "3.5.3",
-    "typescript": "2.3.3",
+    "typescript": "~2.3.3",
     "web-app-manifest-loader": "0.1.1",
-    "webpack": "2.6.1",
-    "webpack-dev-server": "2.4.5",
+    "webpack": "~2.4.0",
```
michael-wirth: downgrading?

src/main/webapp/app/job-search/job-search-filter/job-search-filter.component.ts
```diff
@@ -1,15 +1,16 @@
...
 })
 export class JobSearchFilterComponent implements OnInit {
 
-  constructor() { }
+    constructor() {
+    }
 
-  ngOnInit() {
-  }
+    ngOnInit() {
```
michael-wirth: remove this method if the `OnInit` is not used

src/main/webapp/app/job-search/job-search-filter/job-search-filter.component.ts
```diff
@@ -1,15 +1,16 @@
...
-  templateUrl: './job-search-filter.component.html',
-  styles: []
+    selector: 'jr2-job-search-filter',
+    templateUrl: './job-search-filter.component.html',
+    styles: []
 })
 export class JobSearchFilterComponent implements OnInit {
 
-  constructor() { }
+    constructor() {
```
michael-wirth: empty constructor can also be removed

src/test/javascript/spec/app/job-search/shared/occupation.service.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+            imports: [JobroomTestModule],
+            providers: [
+                OccupationService,
+                { provide: TranslateService, useValue: { currentLang: 'de' } }
+            ]
+        });
+    });
+
+    describe('fetchSuggestions', () => {
+        // GIVEN
```
michael-wirth: GIVEN seems to be at the wrong place. it should be places inside the `it` test. 

Here i'd rather user `PREPARE` or `SETUP`

package.json
birom: It is not downgrading. Caret updates to the most recent minor version.

package.json
birom: Tilde updates to the most recent patch version.

package.json
michael-wirth: ok, we need to document this in the wiki.
otherwise this cause problems in the future

birom: But you are right it is confusing to upgrade with a smaller number... I am going to fix it.

PR #8, opened 2017-07-14
JR2-63: Integrate ElasticSearch Reindexing into frontend

birom:
I have generated the code with the  [generator-jhipster-elasticsearch-reindexer](https://github.com/geraldhumphries/generator-jhipster-elasticsearch-reindexer) template.


src/main/webapp/app/admin/elasticsearch-reindex/elasticsearch-reindex.service.ts
```diff
@@ -0,0 +1,15 @@
...
+
+@Injectable()
+export class ElasticsearchReindexService {
+
+    constructor(
+      private http: Http
+    ) { }
+
+    reindex(): Observable<Response> {
+        return this.http.post('api/elasticsearch/index', {});
```
michael-wirth: the service should call the reindex url of `jobservice` and `referenceservice` -> create new task to reindex a selected microservice

PR #21, opened 2017-07-28
Feature/jr2 96 ui translation exporter importer

birom:


gradle/translation.gradle
```diff
@@ -0,0 +1,137 @@
+apply plugin: "java"
```
michael-wirth: do we really need the "java" plugin for these translation tasks? Remove if not required

src/main/webapp/i18n/de/configuration.json
```diff
@@ -1,10 +1,10 @@
...
         "table": {
-            "prefix": "Präfix",
+            "prefix": "Pr\u00e4fix",
             "properties": "Eigenschaften"
-        }
+        },
+        "title": "Konfiguration"
     }
-}
+}
```
michael-wirth: Is it possible to add am empty line at the end of the json files to satisfy github?

gradle/translation.gradle
birom: No we do not need it. I have removed it.

src/main/webapp/i18n/de/configuration.json
birom: Done

michael-wirth:
Approved.

Is the import/export documented?

PR #33, opened 2017-08-14
JR2-118 Enable gatling tests

varusakov:


.travis.yml
```diff
@@ -54,6 +54,8 @@ deploy:
   on:
     all_branches: true
     condition: "$TRAVIS_BRANCH =~ ^release.*$"
+after_deploy:
```
michael-wirth: disable the gatling test with a commend that there are none test yes, but is planed for later

PR #37, opened 2017-08-18
JR2-119 Change OccupationService search URL

varusakov:


src/main/webapp/app/shared/job-search/service/occupation.service.ts
```diff
@@ -6,7 +6,7 @@ import { TranslateService } from '@ngx-translate/core';
 import { InputType, OccupationAutocomplete, OccupationSuggestion } from './occupation-autocomplete';
 
 const DEFAULT_RESPONSE_SIZE = '10';
-const SEARCH_URL = 'referenceservice/api/_search/occupations';
+const SEARCH_URL = 'referenceservice/api/_search/occupations/synonym';
```
birom: Jhipster uses resource URLs in plural. Would it be possible to follow this convention?
Sorry, this avoided my attention in the other PR. Thank you!

michael-wirth: let's check that on Monday

PR #38, opened 2017-08-21
JR2-109 Create the job detail page

pellenberger:
Tests aren't done yet because the implementation will change (state management).


src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -0,0 +1,109 @@
+<div class="container">
+
+    <div class="breadcrumb">
+        <a routerLink="/job-search">Zurück zu den Suchergebnissen</a>
```
birom: Localization is missing. Please have a look at the "Export / Import translation json files to / from CSV file" section in the README.md.

PR #42, opened 2017-08-22
JR2-50 Implement the workingtimepercentage filter component

pellenberger:
Component is included in the job search state management.


src/main/webapp/app/job-search/job-search-filter/workingtimepercentage-filter/workingtimepercentage-filter.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  private workingtimepercentage$: Observable<Workingtimepercentage>;
+  private min: number;
+  private max: number;
+
+  constructor(private store: Store<JobSearchState>) {
+      this.workingtimepercentage$ = store.select(getFilterWorkingtimepercentage);
+  }
+
+  ngOnInit() {
+      this.workingtimepercentage$.subscribe(
```
birom: If you do not use async pipe you should also unsibscribe.

src/main/webapp/app/job-search/job-search-filter/job-search-filter.component.html
```diff
@@ -16,6 +16,8 @@
     <input type="text" class="form-control form-control-sm" id="unternehmen" placeholder="Name des Unternehmen">
 </div>
 
+<jr2-workingtimepercentage-filter></jr2-workingtimepercentage-filter>
```
birom: I would suggest to name the component to : jr2-working-time-filter (as far as I know there is not such a word that workingtime).

PR #44, opened 2017-08-23
JR2-51: Contract type filter

birom:


src/main/webapp/app/job-search/state-management/state/job-search.state.ts
```diff
@@ -2,6 +2,12 @@ import { TypeaheadMultiselectModel } from '../../../shared/job-search/typeahead-
 import { Job } from '../../../entities/job/job.model';
 import { createFeatureSelector, createSelector } from '@ngrx/store';
 
+export enum ContractType {
+    notDefined,
```
michael-wirth: undefined would be better but that is a keyword.
these are states of the job object.

but looking at it from the ui point of view these are query options. so wouldn't it be better to rename it to 'all'?

birom:
I will create a new pull request instead of this.

PR #50, opened 2017-08-29
JR2-166: Job result list sort ui

birom:


src/main/webapp/app/job-search/state-management/state/job-search.state.ts
```diff
@@ -24,18 +19,13 @@ export interface JobSearchState {
...
-}
-
 export interface JobSearchQuery {
     baseQuery: Array<TypeaheadMultiselectModel>;
     localityQuery: Array<TypeaheadMultiselectModel>;
 }
 
 export interface JobSearchFilter {
-    sort: Sort;
+    sort?: string;
```
michael-wirth: Why don't we use Sort anymore?

PR #52, opened 2017-08-30
JR2-159: Use jhipster's truncateCharacters pipe

birom:


birom:
I've just discovered that jhipster have a truncateCharacters pipe.

src/main/webapp/app/job-search/job-search-list-item/job-search-list-item.component.html
```diff
@@ -16,6 +16,6 @@ <h4 class="text--primary-color"><a [routerLink]="['/job-detail', job.id]">{{job.
         <span class="badge badge-contract" *ngIf="job.permanent">{{'job-search.job-search-list-item.badge.permanent' | translate}}</span>
     </div>
 
-    <p>{{getShortDescription(job.description)}}</p>
+    <p>{{job.description | truncateCharacters: maxDescriptionLength}}</p>
```
michael-wirth: to avoid fixed length truncation we could also use CSS tricks

See https://codepen.io/martinwolf/pen/qlFdp

PR #55, opened 2017-08-30
JR2-177 Show build version information

michael-wirth:


src/main/webapp/app/layouts/version/version.component.ts
```diff
@@ -0,0 +1,27 @@
...
+import { VersionInfo } from './version-info.model';
+
+@Component({
+    selector: 'jr2-version',
+    template: `
+        <div class="version">Version:
+            <span>{{versionInfo.version}}.{{versionInfo.buildNumber}} ({{versionInfo.buildTime | date:'short'}})</span>
+        </div>`,
+    styleUrls: [
+        'version.scss'
```
birom: We usually use URLs with './' prefix in order to avoid problems during AOT compilation.

michael-wirth: fixed

PR #60, opened 2017-09-01
JR2-187 Implement links to the external page and direct application

michael-wirth:


src/main/webapp/app/job-search/job-detail/job-detail.component.ts
```diff
@@ -7,7 +7,10 @@ import 'rxjs/add/operator/switchMap';
 @Component({
     selector: 'jr2-job-detail',
     templateUrl: './job-detail.component.html',
-    styles: []
+    styles: [`.btn-block {
```
pellenberger: Maybe we can discuss about if we want to follow this convention (CSS rules directly in component) or the other (always create a separate file).

PR #61, opened 2017-09-04
Feature/jr2 121 home page

pellenberger:


src/main/webapp/app/home/tools/candidate-search-tool/candidate-search-tool.component.html
```diff
@@ -0,0 +1,133 @@
+<fieldset class="candidate-search">
```
birom: Are the translations missing intentionally? (Be careful this PR also modifes them: https://github.com/alv-ch/jobroom2/pull/62)

pellenberger: Yes it's intentionally. The tools components are not part of this story and as their form and content not clearly defined is, translation will be done later when working on the components.

birom: Thank you!

PR #62, opened 2017-09-04
Jr2 175 use css to truncate description

michael-wirth:


src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -8,27 +8,27 @@
 
         <div class="base__subnavbar" style="font-size: 80%;">
 
-            <div class="logo-box flex-height md-whiteframe-1dp">
+            <div company class="logo-box flex-height md-whiteframe-1dp">
```
birom: Do we need the custom _h4_ and _p_ styles only on the company contact section? If yes, I think "logo-box__company" would be a better option according to the [BEM](https://css-tricks.com/bem-101/) naming style.

src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -8,27 +8,27 @@
...
 
                 <div class="job-detail__content">
-                    <strong>{{'job-detail.contactPerson' | translate}}</strong><br/>
+                    <h4>{{'job-detail.contactPerson' | translate}}</h4>
                     <div *ngIf="job.contact">
-                        {{ job.contact.title }}<br/>
-                        {{ job.contact.firstName }} {{ job.contact.lastName }}<br/>
-                        {{ job.contact.phone }}<br/>
-                        {{ job.contact.email }}<br/>
+                        <p>{{ 'job-detail.contactPerson.salutation.' +job.contact.title |
```
birom: There are some missing traslations here eg.: translation-not-found[job-detail.contactPerson.salutation.Herr]

src/main/webapp/app/job-search/job-search-list-item/job-search-list-item.scss
```diff
@@ -0,0 +1,7 @@
+.description {
```
birom: Have you tried with different browsers? I does not work on my firefox (not the latest: 45.7.0).

src/main/webapp/app/job-search/job-detail/job-detail.component.html
michael-wirth: done

src/main/webapp/app/job-search/job-detail/job-detail.component.html
michael-wirth: did you full reload the page. the json files are cached..

src/main/webapp/app/job-search/job-search-list-item/job-search-list-item.scss
michael-wirth: tested it with chrome (webkit) and firevox (gecko). not tested yet with IE

PR #68, opened 2017-09-06
Jr2 190 remove dummy content on job result list

michael-wirth:


pellenberger:
Please remember to begin your branches name with the prefix _feature/_ :)

PR #69, opened 2017-09-06
Feature/jr2 58 print job

pellenberger:


src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -1,14 +1,17 @@
 <div class="container">
 
-    <div class="breadcrumb">
+    <div class="breadcrumb hidden-print">
```
michael-wirth: `hidden-print` seems to be removed in Bootstrap 4.0.0-beta
https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-beta/css/bootstrap.css

can't we just define all the parts to hide in the print.scss or the component specific scss files? it would be nice if we don't need this boilerplate code in the html

pellenberger: Ok, but I was using the `hidden_print` class defined in the styleguide stylesheet (I didn't know that bootstrap already used it, if it's the case). 
I'm not against moving all the print related CSS in the specific print.scss file. I was wondering myself while coding it. But when I saw those pre-defined classes, I simply used them instead of implementing it again. 
I propose to manually hide all the hidden parts in the SCSS file and remove these print classes from the styleguide stylesheet. Is that ok for you ? 
@birom What do you think ?

michael-wirth: btw. in the beta version there are new classes 'd-print-*'. these classes are used to print block , which will not be visible on screen

pellenberger: Yes it's the same principle as the `hidden-print` class. The styleguide stylesheet already offers such possibilities (e.g. `visible-print-block` class).

birom: I would also hide the unwanted elements in the print css and leave the components as they were before.

pellenberger: Done

PR #75, opened 2017-09-11
JR2-224 Enable https with self signed certificate

varusakov:


src/main/resources/config/application.yml
```diff
@@ -89,6 +89,11 @@ server:
     session:
         cookie:
             http-only: true
+    ssl:
```
michael-wirth: implementation wasn't yet part of the task...

src/main/resources/config/application.yml
```diff
@@ -89,6 +89,11 @@ server:
...
+      key-store: classpath:keystore.p12
```
michael-wirth: this settings are envrionment specific and should be stored in the https://github.com/alv-ch/central-config project and load via `ConfigServer`. Unfortunately it is not possible to load the keystore via `ConfigServer` as it is not possible to submit the user credentials.

src/main/resources/config/application.yml
```diff
@@ -89,6 +89,11 @@ server:
...
+      key-store-password: jobroom
+      key-store-type: PKCS12
+      key-alias: cert
```
michael-wirth: use meaningful names

src/main/resources/config/application.yml
```diff
@@ -89,6 +89,11 @@ server:
     session:
         cookie:
             http-only: true
+    ssl:
+      key-store: classpath:keystore.p12
+      key-store-password: jobroom
```
michael-wirth: password should be encrypted  (via `ConfigServer`)

src/main/resources/config/application.yml
michael-wirth: The `cert` must match at least the CN common name with the server `dev.job-room.ch`

webpack/webpack.dev.js
```diff
@@ -31,7 +31,7 @@ module.exports = webpackMerge(commonConfig({ env: ENV }), {
                 '/v2/api-docs',
                 '/h2-console'
             ],
-            target: 'http://127.0.0.1:8080',
+            target: 'https://127.0.0.1:8080',
```
michael-wirth: The server should now run on port 8443 to indicate this is the `https` port

src/main/resources/config/application.yml
michael-wirth: load the keystore from a centralized location as `https://raw.githubusercontent.com/alv-ch/central-config/master/keystore-development.jks` for example

PR #76, opened 2017-09-12
JR2-225 Enable https

michael-wirth:


src/main/devops/docker-compose.yml
```diff
@@ -3,47 +3,59 @@ services:
     jobroom-app:
         build: dockerfiles/jobroom
         environment:
-            - JAVA_OPTS=-Xmx512m -Xms256m
+            - JAVA_OPTS=-Xmx512m -Xms256m -Djavax.net.ssl.trustStore=/config/truststore.jks -Djavax.net.ssl.trustStorePassword=changeit 
             - SPRING_PROFILES_ACTIVE=development,swagger
-            - SPRING_CLOUD_CONFIG_URI=http://admin:$${jhipster.registry.password}@jhipster-registry:8761/config
-            - SLEEP=30
+            - SPRING_CLOUD_CONFIG_URI=https://admin:$${jhipster.registry.password}@jhipster-registry:8761/config
+#            - SPRING_CLOUD_CONFIG_LABEL=https
```
varusakov: I think we should clean up this commented lines

michael-wirth: true, i enable the line and set the default value, so we just can change the value the next time we have to work with different configurations

PR #79, opened 2017-09-12
JR2-228 Remove classifications from candidate search autocomplete

varusakov:


src/main/webapp/app/shared/job-search/service/occupation.service.ts
```diff
@@ -19,6 +19,31 @@ export class OccupationService {
     }
 
     fetchSuggestions(query: string): Observable<TypeaheadMultiselectModel[]> {
+        return this.fetchSuggestionsInternal(query, (occupationAutocomplete) => {
```
birom: In my opinion this is too complicated. The fetchSuggestionsInternal could return an ```Observable<OccupationAutocomplete>.``` And this can be mapped  e.g.:

```
 fetchSuggestions(query: string): Observable<TypeaheadMultiselectModel[]> {
        return this.fetchSuggestionsInternal(query)
            .map((occupationAutocomplete: OccupationAutocomplete) => {
                const occupations = occupationAutocomplete.occupations
                    .map((o: OccupationSuggestion) => new TypeaheadMultiselectModel(OccupationInputType.OCCUPATION, o.code, o.name, 0));
                const classifications = occupationAutocomplete.classifications
                    .map((c: any) => new TypeaheadMultiselectModel(OccupationInputType.CLASSIFICATION, c.code, c.name, 1));

                return [...occupations, ...classifications];
            });
    }
```

The occupation mapper is dupplicated. We should extract it to a reusable function.

src/main/webapp/app/shared/job-search/service/occupation.service.ts
```diff
@@ -19,6 +19,31 @@ export class OccupationService {
...
+            return [...occupations, ...classifications];
+        });
+    }
+
+    getClassifications(labels: Array<string>): Observable<TypeaheadMultiselectModel[]> {
+        // todo: REST endpoint is needed
+        return Observable.of([]);
+    }
+
+    getOccupations(query: string): Observable<TypeaheadMultiselectModel[]> {
```
birom: Please update the occupation.service.spec.ts test file according to the these chages.

src/main/webapp/app/shared/job-search/service/occupation.service.ts
michael-wirth: (thumbup)

PR #80, opened 2017-09-13
JR2-231: Remove generated job entity resources

birom:


birom:
I fixed the failing tests.

PR #83, opened 2017-09-13
JR2-211 Change favicon

varusakov:


varusakov:
Small review. Only favicon was changed

PR #98, opened 2017-09-21
JR2-255: Refactor job search filter to use enum key pipe

birom:


src/main/webapp/app/home/tools/candidate-search-tool/candidate-search-tool.component.html
```diff
@@ -24,7 +24,7 @@
                 <select class="form-control custom-select">
                     <option selected="" translate="candidate-search.no-selection"></option>
                     <option *ngFor="let experience of experiences | keys"
-                            value="{{experience.value}}"
+                            [ngValue]="experience.value"
```
michael-wirth: Thanks for fixing this!

PR #100, opened 2017-09-21
Feature/jr2 240 candidate search common search params

birom:


src/main/webapp/app/candidate-search/candidate-search-filter/candidate-search-filter.component.html
```diff
@@ -0,0 +1,26 @@
+<p class="text--line-height--h3">{{'candidate-search.filter.title' | translate}}</p>
```
michael-wirth: when should we use attribute style: `translate="candidate-search.filter.title" or interpolate style: `{{'candidate-search.filter.title' | translate}}`

Now we a little mix. Please define a little guideline how to handle translations. Shall we mix them or only use one type?

birom: Ok, I will do that. Thank you!

PR #111, opened 2017-09-27
Feature/jr2 248 language filter

birom:


src/main/webapp/app/candidate-search/language-selector/language-selector.component.ts
```diff
@@ -0,0 +1,116 @@
...
+    removeByIndex(idx: number) {
+        const skills = [...this.selectedLanguageSkills];
+        skills.splice(idx, 1);
+
+        this._onChange(skills);
+        this.writeValue(skills);
+    }
+
+    private _onChange = (_: any) => {
+    }
```
varusakov: What is the goal of this empty function? It's just a placeholder?

src/main/webapp/app/candidate-search/candidate-search-filter/candidate-search-filter.component.ts
```diff
@@ -57,6 +59,13 @@ export class CandidateSearchFilterComponent implements OnInit {
             drivingLicenceCategory: [this.searchFilter.drivingLicenceCategory],
             languageSkills: [[...this.searchFilter.languageSkills]]
         });
+
+        this.filterForm.valueChanges.subscribe((formValue: any) =>
+            console.log(formValue));
```
michael-wirth: is this a debugging left-over? i guess we shouldn't write to console.log

src/main/webapp/i18n/translations.csv
```diff
@@ -53,6 +53,53 @@ candidate-search,candidate-search.graduation.NONE,kein Abschluss,kein Abschluss,
 candidate-search,candidate-search.graduation.NOT_ACCEPTED,Abschluss CH oder ausländisch,Abschluss CH oder ausländisch,Abschluss CH oder ausländisch,Abschluss CH oder ausländisch
 candidate-search,candidate-search.graduation.label,Abschluss,Abschluss,Abschluss,Abschluss
 candidate-search,candidate-search.keywords.label,Suchbegriffe,Suchbegriffe,Suchbegriffe,Suchbegriffe
+candidate-search,candidate-search.language.ar,Arabisch,Arabisch,Arabisch,Arabisch
```
michael-wirth: shouldn't we define the languages only once and define them in a global context instead of defining them in job-details and job-details

we will use them in the candidate-details page again

src/main/webapp/app/candidate-search/language-selector/language-selector.component.ts
birom: This is the default implementation in case the registerOnChange on line 34 is not called.

Similar as the angular's DefaultValueAccessor: https://github.com/angular/angular/blob/9479a106bbc5a3fe99c6d80f12dbb3d23362f22a/packages/forms/src/directives/default_value_accessor.ts

The form module use it: https://github.com/angular/angular/blob/1cfa79ca4e21788e0323baf544704ee7ef7d63ea/packages/forms/src/directives/shared.ts

src/main/webapp/app/candidate-search/candidate-search-filter/candidate-search-filter.component.ts
birom: I would like to validate if the form is correct, but the proper test is in scope of https://alv-ch.atlassian.net/browse/JR2-138. Does it make error on IE?

src/main/webapp/i18n/translations.csv
birom: Ok, I fix it. Thank you!

src/main/webapp/app/candidate-search/candidate-search-filter/candidate-search-filter.component.ts
birom: I removed it.

src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -92,7 +92,7 @@ <h2 class="text--light">{{ job.title }}</h2>
                 <h4 class="job-detail__content__title">{{ 'job-detail.languages' | translate }}</h4>
 
                 <dl class="row" *ngFor="let language of job.languages">
-                    <dt class="col-md-4 text-md-right">{{ 'job-detail.language.' + language.code | translate }}</dt>
+                    <dt class="col-md-4 text-md-right">{{ 'global.reference.language.' + language.code | translate }}</dt>
                     <dd class="col-md-8">
                         <span *ngIf="language.spokenLevel">{{ 'job-detail.language.spoken' | translate }}: {{ 'job-detail.language.level.' + language.spokenLevel | translate}}, </span>
```
michael-wirth: we should also _globalize_ the `job-detail.language.level`

PR #113, opened 2017-09-29
JR2-253 Add Pagination links

Invizible:


src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -103,6 +102,20 @@ <h4 class="job-detail__content__title">{{ 'job-detail.languages' | translate }}<
             <div *ngIf="job.application" class="job-detail__content">
                 <h4 class="job-detail__content__title">{{ 'job-detail.application.title' | translate }}</h4>
 
+                <dl *ngIf="job.application.mailEnabled" class="row">
+                    <dt class="col-md-4 text-md-right">{{ 'job-detail.application.mail' | translate}}</dt>
+                    <dd class="col-md-8">{{ 'job-detail.application.mail.text' | translate}}</dd>
+                </dl>
+                <dl *ngIf="job.application.emailEnabled" class="row">
+                    <dt class="col-md-4 text-md-right">{{ 'job-detail.application.email' | translate}}</dt>
+                    <dd class="col-md-8">{{job.contact?.email || job.company?.email}}</dd>
```
michael-wirth: render with `mailto:` protocol

src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -132,49 +145,56 @@ <h4 class="job-detail__content__title">{{ 'job-detail.application.title' | trans
...
-                        <!--<li class="nav-item">-->
-                        <!--<a class="nav-link" href="#"><i class="fa fa-file-pdf-o"-->
-                        <!--aria-hidden="true"></i>-->
-                        <!--{{'job-detail.action.savePdf' | translate}}</a>-->
-                        <!--</li>-->
-                    </ul>
-                </div>
+                <ul *ngIf="jobNavigationEnabled$ | async" class="pagination pagination--block pagination-alv">
+                    <li class="page-item">
+                        <a class="page-link" (click)="nextJob(true)"><i class="fa fa-chevron-left" aria-hidden="true"></i> Previous</a>
```
michael-wirth: disable the paging if there is no previous or next job and use translations for the button text

src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -103,6 +102,20 @@ <h4 class="job-detail__content__title">{{ 'job-detail.languages' | translate }}<
...
+                    <dd class="col-md-8">{{job.contact?.email || job.company?.email}}</dd>
+                </dl>
+                <dl *ngIf="job.application.phoneEnabled" class="row">
+                    <dt class="col-md-4 text-md-right">{{ 'job-detail.application.phone' | translate}}</dt>
+                    <dd class="col-md-8"><a href="tel:{{job.contact?.phone || job.company?.phone}}">
+                        {{job.contact?.phone || job.company?.phone}}
+                    </a></dd>
+                </dl>
                 <dl class="row">
                     <dt class="col-md-4 text-md-right">{{ 'job-detail.application.hints' | translate}}</dt>
```
michael-wirth: render only if the `application.hints` is not empty

src/main/webapp/app/job-search/job-detail/job-detail.component.ts
```diff
@@ -45,11 +61,20 @@ export class JobDetailComponent implements OnInit, AfterViewInit {
...
+
     private isShowInfoBox(): boolean {
         return this.job.externalId.startsWith('AVAM');
     }
 
     printJob() {
         window.print();
     }
+
+    nextJob(reverse = false): void {
```
michael-wirth: use `nextJob` and `previousJob` - so the code is more readable and easier to understand

src/main/webapp/app/job-search/state-management/effects/job-search.effects.ts
```diff
@@ -73,6 +83,32 @@ export class JobSearchEffects {
...
+
+        if (nextJobIndex === -1) {
+            return new NextJobErrorAction();
+        }
+        const nextJob = state.jobList[nextJobIndex];
+
+        if (nextJob) {
+            return new NextJobLoadedAction({ job: nextJob });
+        } else {
+            this.actions$
```
birom: I find this a little bit confusing. The getNextJob name indicates something other not the LoadNextPageAction.
Instead of that we could do something like this: 
```
 private getNextJob(state: JobSearchState, loadNextJobAction: LoadNextJobAction): Observable<Job> {
        const { currentJobId, reverse } = loadNextJobAction.payload;
        const currentJobIndex = state.jobList.findIndex((element) =>
            element.id === currentJobId);

        const nextJobIndex = currentJobIndex + (reverse ? -1 : 1);

        const nextJob = state.jobList[nextJobIndex];

        if (nextJob) {
            return Observable.of(nextJob);
        } else {
            this.store.dispatch(new LoadNextPageAction());

            return this.actions$
                .ofType(NEXT_PAGE_LOADED)
                .take(1)
                .map((action: NextJobLoadedAction) => action.payload[0])
        }
    }
```

src/main/webapp/app/job-search/state-management/effects/job-search.effects.ts
```diff
@@ -63,6 +65,14 @@ export class JobSearchEffects {
                 .catch((err: any) => Observable.of(new ShowJobListErrorAction(err)))
         );
 
+    @Effect()
```
birom: In my opinion the navigation should be here instead of the JobDetailComponent. 
eg.:

```
 @Effect()
    loadNextJob$: Observable<Action> = this.actions$
        .ofType(LOAD_NEXT_JOB)
        .distinctUntilChanged()
        .withLatestFrom(this.store.select(getJobSearchState))
        .switchMap(([action, state]) => this.getNextJob(state, action as LoadNextJobAction))
        .map((job: Job) => job
            ? new NextJobLoadedAction({ job })
            : new NextJobErrorAction());

    @Effect({ dispatch: false })
    nextJobLoaded$: Observable<Action> = this.actions$
        .ofType(NEXT_JOB_LOADED)
        .do((action: NextJobLoadedAction) => {
            const job = action.payload.job;
            this.router.navigate(['/job-detail', job.id])
        });
```

src/main/webapp/app/job-search/job-detail/job-detail.component.html
Invizible: I'm not sure if I can do disabling for the next button, because if you ran out of jobs, saved on the front-end, it will load the next page (20 new jobs) and show the next one.

birom: We have the totalJobCount in the JobSearchState so it is possible, but it agree that it makes the implementation too complicated. What about showing the first job again if we reach the end of the list?

Invizible: I don't thing that's logical, because when user scrolls down, he gets new batch of jobs, so I think it would be logical to give him the same experience but in a different view

Invizible: yep, I can use totalJobCount, I was thinking about situation, when user gets to the last job, and next button is disabled forever even if new job is added on the server, unless he reloads the page, or tries to scroll down again, but agree that nobody will loop through all jobs :) so it's not a case in my opinion

birom: I meant when the user already scrolled  the last page, so there are no more pages left.
But I also think that we do not need this.

src/main/webapp/app/job-search/state-management/effects/job-search.effects.ts
Invizible: Thanks, good suggestion

src/main/webapp/app/job-search/state-management/effects/job-search.effects.ts
Invizible: cool, thanks

src/main/webapp/app/job-search/job-detail/job-detail.component.ts
```diff
@@ -14,29 +24,37 @@ import { Job } from '../services';
...
-    async: any;
+    private unsubscribe$: Subject<void> = new Subject<void>();
 
-    constructor(private route: ActivatedRoute, private referenceService: ReferenceService) {
+    constructor(private route: ActivatedRoute,
+                private referenceService: ReferenceService,
+                private store: Store<JobSearchState>) {
+        this.jobNavigationEnabled$ = this.store.select(getJobNavigationEnabled);
+        this.store.select(getSelectedJobIndex)
+            .takeUntil(this.unsubscribe$)
```
birom: Until now we have done something like this:
```
this.subscription = this.store.select(getJobListScrollY).subscribe((a)=> this.x=a)
```
and than in the ``ngOnDestroy()``
```
if (this.subscription) {
            this.subscription.unsubscribe();
        }
```

Is this solution better than that?

src/main/webapp/app/job-search/job-detail/job-detail.resolver.ts
```diff
@@ -8,15 +8,27 @@ import {
...
     resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Job> {
-        return this.jobService.find(route.params['id'])
+        const jobId = route.params['id'];
+
+        return this.store.select(getJobList)
+            .take(1)
+            .map((jobs: Array<Job>) =>
+                jobs.find((jobItem) => jobItem.id === jobId))
+            .flatMap((storedJob: Job) => storedJob
+                ? Observable.of(storedJob)
```
birom: We have a small problem here:
When the items in the hit list are highlighted, the detailed view contains the escaped highlight text eg.:
```
<em>Java</em>/<em>JavaEE</em>-EntwicklerIn
``` 
We could change the REST endpoint and to add a normal and a highlighted version of the title and the description.
I suggest to do that in the next sprint and use the original version without the caching.

src/main/webapp/app/job-search/job-detail/job-detail.component.ts
```diff
@@ -45,11 +63,25 @@ export class JobDetailComponent implements OnInit, AfterViewInit {
         this.showInfoBox = this.isShowInfoBox();
     }
 
+    ngOnDestroy(): void {
+        this.unsubscribe$.next();
```
birom: See comment at line 44.

src/main/webapp/app/job-search/state-management/effects/job-search.effects.ts
```diff
@@ -63,6 +68,25 @@ export class JobSearchEffects {
...
+        .distinctUntilChanged()
+        .withLatestFrom(this.store.select(getJobSearchState))
+        .switchMap(([action, state]) =>
+            this.getNextJob(state, action as LoadNextJobAction | LoadPreviousJobAction))
+        .map((selectedJob: SelectedJob) => selectedJob
+            ? new NextJobLoadedAction(selectedJob)
+            : new NextJobErrorAction());
+
+    @Effect({ dispatch: false })
+    nextJobLoaded$: Observable<Action> = this.actions$
```
birom: Please write some test for this in the [job-search.effects.spec.ts](https://github.com/alv-ch/jobroom2/blob/develop/src/test/javascript/spec/app/job-search/state-management/effects/job-search.effects.spec.ts) . We use [marbles](https://github.com/ReactiveX/rxjs/blob/master/doc/writing-marble-tests.md) to test observables.
Thank you!

src/main/webapp/app/job-search/job-detail/job-detail.component.ts
Invizible: I think yes, when you have many subscriptions you need to write ugly code:
```
this.subscription.add(this.store.select().subscribe()) 
this.subscription.add(this.store.select().subscribe()) 
this.subscription.add(this.store.select().subscribe()) 
```
So my version is more clear
BTW, here is the link where I found this solutions https://stackoverflow.com/a/41177163/2367029

birom: Ok, thank you!

Invizible: But now, I have only 1 subscription, so I can replace this solution with that one you suggest

src/main/webapp/app/job-search/state-management/effects/job-search.effects.ts
Invizible: Sure

src/main/webapp/app/job-search/job-detail/job-detail.component.ts
birom: It is up to you, I usually try to avoid subscribe and use the async pipe (with smart and presentation components) instead of that.

src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -104,7 +103,23 @@ <h4 class="job-detail__content__title">{{ 'job-detail.languages' | translate }}<
             <div *ngIf="job.application" class="job-detail__content">
                 <h4 class="job-detail__content__title">{{ 'job-detail.application.title' | translate }}</h4>
 
-                <dl class="row">
+                <dl *ngIf="job.application.mailEnabled" class="row">
```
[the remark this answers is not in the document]
Invizible: Ok, I can use divs with classes from styleguide

PR #126, opened 2017-10-05
JR2-273 Add sticky footer and add min-height to the cover image

michael-wirth:
implemented according https://css-tricks.com/couple-takes-sticky-footer/


src/main/webapp/app/home/home.scss
```diff
@@ -12,19 +11,20 @@ Main page styles
...
-only screen and (        min-device-pixel-ratio: 2),
-only screen and (                min-resolution: 192dpi),
-only screen and (                min-resolution: 2dppx) {
+@media only screen and (-webkit-min-device-pixel-ratio: 2),
+only screen and (min--moz-device-pixel-ratio: 2),
+only screen and (-o-min-device-pixel-ratio: 2/1),
+only screen and (min-device-pixel-ratio: 2),
+only screen and (min-resolution: 192dpi),
+only screen and (min-resolution: 2dppx) {
     .hipster {
```
birom: Do we still need this class?

src/main/webapp/content/scss/global.scss
```diff
@@ -19,3 +19,29 @@ Print configuration
 button.close {
     cursor: pointer;
 }
+
```
birom: Is this just temporary solution, until the migration to the new styleguide is done?

src/main/webapp/app/home/home.scss
michael-wirth: no - but I didn't change anything there - it's just reformatting issue

michael-wirth: we once will need a cleanup issue to remove that code - or we reuse it for jobroom somehow

birom: Ok, I see. Thank you!

src/main/webapp/content/scss/global.scss
michael-wirth: so far this will be the permanent solution - we need to discuss the solution with thomas and if we put some css into the styleguide

PR #130, opened 2017-10-05
JR2-309 toolbar for candidate result page

Invizible:


src/main/webapp/app/job-search/job-search-toolbar/job-search-toolbar.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="toolbar toolbar--light md-whiteframe-3dp sticked--top sticked-item">
     <div class="container">
-        <form [formGroup]="toolbarForm" (ngSubmit)="search(toolbarForm.value)" novalidate
+        <form #form [formGroup]="toolbarForm" (ngSubmit)="search(toolbarForm.value)" novalidate
```
birom: Is the template ref. variable used somewhere?

Invizible: I don't need it, will remove

src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.html
```diff
@@ -0,0 +1,34 @@
...
+                                           [itemLoader]="fetchLocalitySuggestions"
+                                           [editable]="false"
+                                           [focusFirst]="true"
+                                           [placeHolder]="'candidate-search.toolbar.residence.placeholder' | translate"></jr2-typeahead-multiselect>
+            </div>
+
+            <div class="form-group form-group--wide">
+                <label for="graduation" class="sr-only">{{ 'candidate-search.toolbar.graduation.label' | translate}}</label>
+                <select formControlName="graduation" class="form-control form-control-lg custom-select" id="graduation">
+                    <option selected="" value="null" translate="candidate-search.toolbar.graduation.default"></option>
```
michael-wirth: I guess the value should be empty and not "keinen Abschluss ausgewählt"

I'll ask Thomas...

src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.scss
```diff
@@ -0,0 +1,9 @@
+.sticky-top {
```
michael-wirth: align layout with job-search-toolbar. At the moment the styleguide has some major changes, which don't work properly. Please create a JIRA task to review the layouts with the styleguide as soon as it is fixed. 

If there is a quick solution you can also provide a proposal for the layout

src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.html
michael-wirth: Got answer from Thomas. Use `Abschluss auswählen` instead of *keinen Abschluss ausgewählt*

src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.html
```diff
@@ -0,0 +1,34 @@
+<div class="toolbar toolbar--light md-whiteframe-3dp sticky-top">
+    <div class="container">
+        <form [formGroup]="toolbarForm" (ngSubmit)="search(toolbarForm.value)" novalidate
+              class="form-inline toolbar-form job-search__form">
+
+            <div class="form-group form-group--wide">
+                <jr2-typeahead-multiselect formControlName="occupation"
```
birom: There are some differences between the candidate search toolbar and the job search toolbar.
This should be a ng-bootstrap typehaed (see: home/tools/candidate-search-tool).
Do you want to open an other PR for that?

Invizible: I can do that in this PR

birom: Ok, thank you!

src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.html
Invizible: so we will have 'Abschluss auswählen' and then other options would be: CH-Abschluss, CH-Abschluss oder CH anerkannt, Abschluss CH oder ausländisch, kein Abschluss.
Is it correct?

src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.html
Invizible: As far as I see it allows to select only one option

varusakov: We need single selection here.

Invizible: ok, another question, where is the 'skills' field, I removed it from the search filter, because it does not present in the styleguide. I thought this one 'occupation' field will work for both, but if we have only one selection I don't understand how I can find 'java, cool developer' candidate :)

birom: I think the styleguide is not complete. We had a discussion with Thomas when we implemented the the landing page. The occupation should be a single select (contrary to the job-search page), and the different input type should emphasize this difference. The skill input should be a MultiselectComponent (https://github.com/alv-ch/jobroom2/tree/develop/src/main/webapp/app/shared/input-components/multiselect).

Invizible: Ok, I will change occupation field and later we will do something with skills

birom: Thank you and sorry for the confusion. There is one more thing here.. 
In the last sprint the residence was a simple select component see.: candidate-search-tool.component.html line 22 and candidate-search-filter.component.html line 37. We discussed that this should be changed to a multiselect. But there are a few of things to clarify here. 
For this reason I suggest to use a simple html component here and we will change it to a multiselect later.

Invizible: ok, will do

PR #133, opened 2017-10-06
JR2-339 Redesign candidate search landing page

michael-wirth:


src/main/webapp/i18n/translations.csv
```diff
@@ -312,6 +301,8 @@ home,home.tools.candidateSearchTab,Kandidaten finden,Kandidaten finden,Kandidate
 home,home.tools.job-search.search-button.title,Traumjob finden!,Traumjob finden!,Traumjob finden!,Traumjob finden!
 home,home.tools.jobPublicationTab,Job publizieren,Job publizieren,Job publizieren,Job publizieren
 home,home.tools.jobSearch.title,Finden Sie jetzt Ihren Traumjob!,Finden Sie jetzt Ihren Traumjob!,Finden Sie jetzt Ihren Traumjob!,Finden Sie jetzt Ihren Traumjob!
+home,home.tools.candidate-search.title,Finden Sie passende Kandidaten,Finden Sie passende Kandidaten,Finden Sie passende Kandidaten,Finden Sie passende Kandidaten
```
michael-wirth: we need to define an unique message key format
- shall we use kebab-case or camelCase?
- when do we use `title`, `label` or `placeholder`? Can they change depending on the the responsive design? or do we maintain them twice if its a label and in another view a placeholder?

src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.html
```diff
@@ -13,19 +13,22 @@
             </div>
             <div class="form-group form-group--wide">
                 <select formControlName="residence" class="form-control form-control-lg custom-select">
-                    <option selected="" value="null" translate="candidate-search.toolbar.residence.placeholder"></option>
-                    <option *ngFor="let greaterRegion of greaterRegions | keys" [ngValue]="greaterRegion.key">
-                        {{'global.reference.greaterRegion.' + greaterRegion.value | translate}}
+                    <option selected="" value="null"
```
michael-wirth: how shall we deal with this custom mockups of Thomas?

Shouldn't we stick closer to the bootstrap defaults? Should be use  here just the `form-inline` format as in https://getbootstrap.com/docs/4.0/components/forms/#inline-forms?

src/main/webapp/i18n/translations.csv
birom: I vote for kebab-case. I think for the inputs we need separate label and placeholder texts. I used title for the form group titles or tab titles.

PR #143, opened 2017-10-18
JR2-318: Candidate search list

birom:


src/main/webapp/app/candidate-search/state-management/effects/candidate-search.effects.ts
```diff
@@ -0,0 +1,90 @@
...
+            this.candidateService.search(toSearchRequest(action as SearchCandidatesAction, state))
+                .map(toCandidateProfileListLoadedAction)
+                .catch((err: any) => Observable.of(new ShowCandidateListErrorAction(err)))
+        );
+
+    @Effect({ dispatch: false })
+    selectCandidateProfile$: Observable<Action> = this.actions$
+        .ofType(SELECT_PROFILE_CANDIDATE)
+        .do((action: SelectCandidateProfileAction) =>
+            this.router.navigate(['/candidate-search', action.payload.profile.id]));
```
Invizible: you can inline this navigation into component template, I do not use index anymore for pagination buttons, there is no need of this extra action

birom: I changed it

PR #145, opened 2017-10-18
JR2-345 Implement mapping of filter parameters to url params

varusakov:
Need to investigate why preferred work canton/locality not working


src/main/webapp/app/candidate-search/state-management/util/search-request-mapper.ts
```diff
@@ -1,13 +1,63 @@
...
+    const residence = mapResidence(searchFilter.residence);
+    const workLoad = mapWorkLoad(searchFilter.workload);
+    const languageSkills = mapLanguageSkills(searchFilter.languageSkills);
+
     return {
-        occupation: occupation ? occupation.code : '',
+        occupation: occupation ? occupation.code : null,
+        skills: searchFilter.skills,
+        experience: Experience[experience],
+        'residence': residence,
```
birom: You can use instead of _'residence': residence_, just simple _residence,_  (same applys for the workload and languageSkills too).

src/main/webapp/app/candidate-search/services/candidate-search-request.ts
```diff
@@ -1,23 +1,31 @@
...
     page: number,
     size: number,
 }
 
+export interface WorkLoad {
+    min: number
+    max: number
+}
+
+export interface CandidateLanguageSkill {
```
michael-wirth: Why do we need a `CandidateLanguageSkill`?

src/main/webapp/app/candidate-search/state-management/util/search-request-mapper.ts
michael-wirth: agree, use unique formats

varusakov: In that case compiler throws warning 'Expected property shorthand in object literal'

michael-wirth: only for `residence` property?

src/main/webapp/app/candidate-search/services/candidate.service.ts
```diff
@@ -12,6 +15,34 @@ export class CandidateService {
     private resourceUrl = 'candidateservice/api/candidates';
     private searchUrl = 'candidateservice/api/_search/candidates';
 
+    private static isLanguageSkill(arg: any): arg is CandidateLanguageSkill {
+        return arg.code !== undefined;
```
michael-wirth: does this mean if the object has a property `code` it is a CandidateLanguageSkill?

src/main/webapp/app/candidate-search/services/candidate.service.ts
```diff
@@ -12,6 +15,34 @@ export class CandidateService {
...
+                params.append(key, v);
+            }
+        });
+    }
+
+    private static convertObjectToParams(params: URLSearchParams, obj: any, key: string, index?: number): void {
+        for (const property in obj) {
+            if (obj.hasOwnProperty(property)) {
+                let param: string;
+                if (CandidateService.isLanguageSkill(obj)) {
```
michael-wirth: This whole mapping should be simpler or needs some documentation

src/main/webapp/app/candidate-search/services/candidate-search-request.ts
```diff
@@ -1,23 +1,31 @@
...
-    residence: string,
-    availability: Availability,
-    workload: number,
-    workSystem: WorkForm,
-    educationLevel: ISCED_1997,
-    graduation: Graduation,
-    drivingLicenceCategory: DrivingLicenceCategory
-    languageSkills: Array<LanguageSkill>,
+    occupation?: string,
+    skills?: Array<string>,
```
michael-wirth: skills must be URL encrypted. See bug JR2-349. Check possible solutions with @birom

src/main/webapp/app/candidate-search/services/candidate-search-request.ts
varusakov: We can't use `LanguageSkill` from `shared-types.ts` as it uses `CEFR_Level` enum for `spoken` and `written` properties. We need string representation of that properties in request parameters.

src/main/webapp/app/candidate-search/services/candidate.service.ts
varusakov: Yes, you are right. We can't use typescript `instanceof` operator on interfaces to check type of object.

varusakov: Thats why I wrote this function. But we can change `CandidateLanguageSkill` to class and use more strict `instanceof`

src/main/webapp/app/candidate-search/state-management/util/search-request-mapper.ts
varusakov: For any property which value defined as variable. Possible solution to assign value directly:
`residence: mapResidence(searchFilter.residence)`

src/main/webapp/app/candidate-search/services/candidate.service.ts
varusakov: REST service requires parameters in different format:
* `List<string>` - param=1&param=2
* `object` - objectName.propertyName=value
* `List<object>`- objectName[index].propertyName=value

That's why it so complicated

PR #149, opened 2017-10-20
JR2-361 Hide work period

Invizible:
As discussed previously with @michael-wirth we do not have data for work period, so this component can't be finished for now


src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.html
```diff
@@ -39,7 +39,7 @@ <h2 class="text--light">Title</h2>
                     {{ 'global.reference.canton.' + candidateProfile.residenceCantonCode | translate}}
                 </span>
                 <span class="badge badge-gray">{{ candidateProfile.workLoad }}%</span>
-                <jr2-work-period [jobExperiences]="candidateProfile.jobExperiences"></jr2-work-period>
+                <!--<jr2-work-period [jobExperiences]="candidateProfile.jobExperiences"></jr2-work-period>-->
```
michael-wirth: remove the comment and the `jr2-work-period` component (I guess it's not used at all yet). 
this field will be replaced by `permanent` field (befristed/unbefristed), but this property is not yet part of the candidate model

PR #151, opened 2017-10-20
Feature/jr2 345 create mapper for candidate search request

varusakov:


src/main/webapp/app/candidate-search/state-management/util/search-request-mapper.ts
```diff
@@ -13,33 +13,41 @@ import {
...
         experience: Experience[experience],
-        'residence': residence,
+        residence: mapResidence(searchFilter.residence),
         availability: Availability[availability],
-        'workLoad': workLoad,
+        workLoad: mapWorkLoad(searchFilter.workload),
         workForm: WorkForm[workForm],
         educationLevel: ISCED_1997[educationLevel],
-        graduation: Graduation[graduation],
+        graduation: mapGraduation(graduation),
```
michael-wirth: Why is `Graduation` handled differently than the other enumerations?

varusakov: It's passed differently. It can be passed as string or number

michael-wirth: but why is it passed once as string and once as number? I'd prefer if it we wouldn't mix the values if possilbe

varusakov: I'll try to fix that

michael-wirth: Are you still working on this?

PR #165, opened 2017-10-24
JR2-399 Order cantons and regions alphabetically

varusakov:


src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.ts
```diff
@@ -101,6 +106,22 @@ export class CandidateDetailComponent implements OnInit {
...
+            'global.reference.region.');
+        this.preferredWorkCantons$ = this.translateValues(candidate.preferredWorkCantons,
+            'global.reference.canton.');
+    }
+
+    private translateValues(values: Array<string>, keyPrefix: string): Observable<Array<string>> {
+        if (values && values.length > 0) {
+            const keys = values.map((value) => keyPrefix.concat(value));
+            return this.translateService.stream(keys)
+                .map((translations) => keys.map((key) => translations[key]).sort());
```
Invizible: maybe we can just write sort pipe or use some exististing library, and keep this component much simplier

src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.ts
```diff
@@ -101,6 +106,22 @@ export class CandidateDetailComponent implements OnInit {
...
 
+    private populatePreferredWorkLocations(candidate: CandidateProfile): void {
+        this.preferredWorkRegions$ = this.translateValues(candidate.preferredWorkRegions,
+            'global.reference.region.');
+        this.preferredWorkCantons$ = this.translateValues(candidate.preferredWorkCantons,
+            'global.reference.canton.');
+    }
+
+    private translateValues(values: Array<string>, keyPrefix: string): Observable<Array<string>> {
+        if (values && values.length > 0) {
```
Invizible: you can omit '> 0'  'values && values.length' is enough for not empty check

src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.ts
Invizible: actually Jhipster comes with one : https://github.com/jhipster/ng-jhipster/blob/master/src/pipe/order-by.pipe.ts

michael-wirth: agree, we should wrap the sorting function... if there is an existing reusable solution we should go for that

varusakov: ok, I'll use JHipster pipe for ordering but translation of values will be in component

src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.ts
varusakov: Thanks, fixed

src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.ts
```diff
@@ -61,6 +63,9 @@ export class CandidateDetailComponent implements OnInit {
             .map((candidateProfile) => candidateProfile.id)
             .flatMap((id) => this.candidateService.findCandidate(id));
 
+        this.candidateProfile$
```
birom: Trasnlating a collection is also a common task. Could you please write a todo here, we should think about creating a pipe for that too in the future. 
Please map the candidateProfile$ directly to preferredWorkRegions$ and preferredWorkCantons$ instead of using a subscribe.

src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.ts
```diff
@@ -101,6 +106,22 @@ export class CandidateDetailComponent implements OnInit {
...
     }
 
+    private populatePreferredWorkLocations(candidate: CandidateProfile): void {
+        this.preferredWorkRegions$ = this.translateValues(candidate.preferredWorkRegions,
+            'global.reference.region.');
+        this.preferredWorkCantons$ = this.translateValues(candidate.preferredWorkCantons,
+            'global.reference.canton.');
+    }
+
+    private translateValues(values: Array<string>, keyPrefix: string): Observable<Array<string>> {
```
birom: Trasnlating a collection is also a common task. Could you please write a todo here, we should think about creating a pipe for that too in the future.

PR #181, opened 2017-10-29
JR2-423 Show male/female occupation labels

michael-wirth:


webpack/webpack.mixed.js
```diff
@@ -0,0 +1,106 @@
...
+const BrowserSyncPlugin = require('browser-sync-webpack-plugin');
+const WebpackNotifierPlugin = require('webpack-notifier');
+const path = require('path');
+
+const utils = require('./utils.js');
+const commonConfig = require('./webpack.common.js');
+
+const ENV = 'development';
+
+module.exports = webpackMerge(commonConfig({env: ENV}), {
```
birom: Is it not possible to extend the webpack.dev.js with the proxy config, instead of dupplicate the whole configuration?
(We also did that in webpack.development.js file)

src/main/webapp/app/candidate-search/candidate-search-list-item/candidate-search-list-item.component.ts
```diff
@@ -39,9 +39,7 @@ export class CandidateSearchListItemComponent implements OnInit {
 
         if (relevantJobExperience) {
             this.jobExperience$ = this.occupationService.findOccupationByCode(relevantJobExperience.occupationCode)
-                .map((occupation: Occupation) => Object.assign({}, relevantJobExperience, { occupationLabels: occupation.labels }))
-                .combineLatest(this.currentLanguage$)
-                .map(([experience, lang]) => Object.assign({}, experience, { occupation: experience.occupationLabels[lang] }));
+                .map((occupation: Occupation) => Object.assign({}, relevantJobExperience, { occupation: occupation.labels.male + ' / ' + occupation.labels.female }));
```
birom: Please also remove the _private currentLanguage$: Observable<string>;_ and its initialisation in the constuctor.

webpack/webpack.mixed.js
michael-wirth: yes that would be great, but that needs more expertise
I'll create a task to clean it up, it's just a template to run some microservices locally

PR #182, opened 2017-10-30
Feature/upgrade styleguide

tdascoli:
Upgrade to alv-styleguide v0.0.11
Fix JR2-271
Fix JR2-274
Fix JR2-380

### DIFF
--- a/src/main/webapp/content/scss/global.scss
+++ b/src/main/webapp/content/scss/global.scss
@@ -73,3 +73,13 @@ footer,
         display: none;
     }
 }
+
+/*=================================================================================
+Toolbar (alv-styleguide v0.0.11)
+=================================================================================== */
+
+.sticked--top .toolbar-form,
+.sticky-top .toolbar-form {
+    margin-top: 0;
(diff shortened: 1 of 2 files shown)


src/main/webapp/content/scss/global.scss
```diff
@@ -73,3 +73,13 @@ footer,
         display: none;
     }
 }
+
```
birom: Do we still need the everything between line 19 and line 47, or is something also available in the new styleguide?
```
button.close {
    cursor: pointer;
}

html, body {
    margin: 0px;
}

div.main {
    min-height: 100vh;
    margin-bottom: -61px;
}

div.main:after {
    content: "";
    display: block;
}

footer,
.footer,
div.main:after {
    height: 61px;
}

footer,
.footer {
    bottom: 0;
    width: 100%;
}
```

tdascoli: for the moment yes.

birom: Thank you!

PR #190, opened 2017-10-31
Feature/optimizing landingpages

tdascoli:
Optimizing the landingpages for job-search and candidate-search to fit for tablet/mobile view.

### DIFF
--- a/src/main/webapp/app/home/tools/candidate-search-tool/candidate-search-tool.component.scss
+++ b/src/main/webapp/app/home/tools/candidate-search-tool/candidate-search-tool.component.scss
@@ -1,12 +1,3 @@
-h1 {
-    font-weight: 300;
-    padding-bottom: 1rem;
-}
-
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/home/tools/candidate-search-tool/candidate-search-tool.component.scss
```diff
@@ -1,11 +1,4 @@
-h1 {
-    font-weight: 300;
-    padding-bottom: 1rem;
-}
-
-.tab-content--job-room {
-    padding: 2rem;
-}
+@import '../../../../content/scss/jbr2_variables.scss';
```
varusakov: Do we use variables from this imported file?

tdascoli: no, but without i get a weird error. i'm still investigating in it.

PR #205, opened 2017-11-06
JR2-453 Migrate to Jhipster 4.10.2

michael-wirth:


src/main/webapp/app/polyfills.ts
```diff
@@ -1,5 +1,69 @@
...
+/**
+ * Required to support Web Animations `@angular/animation`.
+ * Needed for: All but Chrome, Firefox and Opera. http://caniuse.com/#feat=web-animation
+ **/
+// import 'web-animations-js';  // Run `npm install --save web-animations-js`.
+
+/***************************************************************************************************
+ * Zone JS is required by Angular itself.
+ */
+import 'zone.js/dist/zone';  // Included with Angular CLI.
```
birom: @pellenberger added zonejs to the index.html  in case of IE (see.: https://github.com/alv-ch/jobroom2/commit/78280efab733f4c68628ae927dcefb91a87dc352). Do we still need that?

PR #206, opened 2017-11-06
JR2-404 Add zip and city input fields for locations outside of Switzerland

Invizible:


src/main/webapp/app/home/tools/job-publication-tool/zip-code/zip-code.component.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    ngOnInit() {
+        const validators = this.optional ? [] : [Validators.required];
+        this.formGroup.addControl(this.controlName, this.fb.group({
+            zip: ['', [...validators, Validators.pattern(/^[0-9]*$/)]],
+            city: ['', validators]
+        }));
+
+        this.countryControl.valueChanges
```
birom: I think we could map it to a switzSelected$ Observable and spare the ChangeDetectorRef.markForCheck and the ngOnDestroy. Or do you need them at other places too?

Invizible: thanks, good suggestion

PR #209, opened 2017-11-07
JR2-453: Drop the database on the dev environment

birom:
The migraion to the latest jhipster version caused the following error during deployment:
```
dev_jobroom-app.1.0q8p8siwj16d@jobroom2-dev    | org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [ch/admin/seco/jobroom/config/DatabaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.exception.ValidationFailedException: Validation Failed:
dev_jobroom-app.1.0q8p8siwj16d@jobroom2-dev    |      1 change sets check sum
dev_jobroom-app.1.0q8p8siwj16d@jobroom2-dev    |           config/liquibase/changelog/00000000000000_initial_schema.xml::00000000000001::jhipster was: 7:d2190cf8eff6a01816c8cbc8f5e48e59 but is now: 7:602584c8eec728d076d3e51bf420cc45
```


birom:
We fixed this issue using an other solution.

PR #217, opened 2017-11-08
Replace <i> and <b> tags

tdascoli:


pellenberger:
Shouldn't these kinds of change be done in the styleguide ?

tdascoli:
This has to be done in both repos.

PR #220, opened 2017-11-09
JR2-425 Create custom RouterStateSerializer

pellenberger:


src/main/webapp/app/shared/custom-router-state-serializer/custom-router-state-serializer.ts
```diff
@@ -0,0 +1,32 @@
...
+
+        // Only return an object including the URL and query params
+        // instead of the entire snapshot
+        return { url, queryParams };
+    }
+}
+
+export const reducers: ActionReducerMap<State> = {
+    routerReducer: routerReducer
+};
```
birom: Please add a newline here.

PR #228, opened 2017-11-15
Jr2 455 create import user endpoint

michael-wirth:


src/main/java/ch/admin/seco/jobroom/web/rest/UserResource.java
```diff
@@ -149,9 +150,31 @@ public UserResource(UserRepository userRepository, UserService userService, Mail
...
+
+    }
 
+    @PatchMapping("/users")
+    @Timed
+    @Secured(AuthoritiesConstants.ADMIN)
+    public ResponseEntity<UserDTO> importUser(@Valid @RequestBody ManagedUserVM managedUserVM) throws URISyntaxException {
+        log.debug("REST request to import User : {}", managedUserVM);
+        Optional<User> existingUser = userRepository.findOneByEmailIgnoreCase(managedUserVM.getEmail());
+        if (existingUser.isPresent() && (!existingUser.get().getLogin().equals(managedUserVM.getLogin()))) {
```
varusakov: `LoginAlreadyUsed` is unchecked exception. I think we could refactor this check in such way: `existingUser.filter(user -> !user.getLogin().equals(managedUserVM.getLogin())).ifPresent((user) -> throw new LoginAlreadyUsedException())`

src/main/java/ch/admin/seco/jobroom/web/rest/UserResource.java
```diff
@@ -149,9 +150,31 @@ public UserResource(UserRepository userRepository, UserService userService, Mail
...
+            throw new LoginAlreadyUsedException();
+        }
+        if (!existingUser.isPresent()) {
```
varusakov: Could we use `map().orElseGet()` chain here? For example:
```java
existingUser.map(user -> {
    managedUserVM.setId(user.getId());
    Optional<UserDTO> updatedUser = userService.updateUser(managedUserVM);
    return ResponseUtil.wrapOrNotFound(updatedUser, 
        HeaderUtil.createAlert("userManagement.updated", managedUserVM.getLogin()));
}).orElseGet(() -> {
    User newUser = userService.createUser(managedUserVM);
    return ResponseEntity.created(new URI("/api/users/" + newUser.getLogin()))
                 .headers(HeaderUtil.createAlert("userManagement.created", newUser.getLogin()))
                 .body(new UserDTO(newUser));
});
```

src/main/resources/config/application-dev.yml
```diff
@@ -19,6 +19,7 @@ eureka:
     client:
         service-url:
             defaultZone: http://admin:${jhipster.registry.password}@localhost:8761/eureka/
+        enabled: false
```
varusakov: Why this was disabled?

src/main/java/ch/admin/seco/jobroom/service/UserService.java
```diff
@@ -118,6 +124,9 @@ public User registerUser(ManagedUserVM userDTO) {
         newUser.setActivationKey(RandomUtil.generateActivationKey());
         authorities.add(authority);
         newUser.setAuthorities(authorities);
+        if (nonNull(userDTO.getOrganizationId())) {
+            newUser.setOrganization(organizationRepository.findByExternalId(userDTO.getOrganizationId()).get());
```
varusakov: I think that calling `Optional.get()` is bad idea. It could be replaced with:
```java
organizationRepository.findByExternalId(userDTO.getOrganizationId())
    .ifPresent(newUser::setOrganization)
```

src/main/java/ch/admin/seco/jobroom/service/UserService.java
```diff
@@ -147,6 +156,9 @@ public User createUser(UserDTO userDTO) {
         user.setResetKey(RandomUtil.generateResetKey());
         user.setResetDate(Instant.now());
         user.setActivated(true);
+        if (nonNull(userDTO.getOrganizationId())) {
+            user.setOrganization(organizationRepository.findByExternalId(userDTO.getOrganizationId()).get());
```
varusakov: Same as for comment above

varusakov:
I think we also need to provide tests for `UserResource`. At least in another PR

src/main/resources/config/application-dev.yml
michael-wirth: changed it for testing. i reverted the change

src/main/java/ch/admin/seco/jobroom/web/rest/UserResource.java
michael-wirth: hmm... this is not very readable code either and won't work as new URI is throwing an expected exception

michael-wirth:
will write tests in another pull request. hope i have fixed all the findings now

PR #230, opened 2017-11-15
JR2-476 Change date format on job publication tool

varusakov:


src/main/webapp/app/shared/util/ngb-date-moment-parser-formatter.ts
```diff
@@ -0,0 +1,31 @@
...
+        const date = moment(value, DATE_FORMAT);
+        return date.isValid() ? {
+            year: date.year(),
+            month: date.month() + 1,
+            day: date.date()
+        } : null;
+    }
+
+    format(dateStruct: NgbDateStruct): string {
+        if (dateStruct === null) {
```
Invizible: you can write: if (!dataStruct)

package.json
```diff
@@ -41,7 +41,8 @@
     "rxjs": "5.4.2",
     "swagger-ui": "2.2.10",
     "tether": "1.4.0",
-    "zone.js": "^0.8.18"
+    "zone.js": "^0.8.18",
+    "moment": "^2.19.2"
```
michael-wirth: do we really need momen
js for a simple date formatting? is the angular date formater not fitting?

varusakov: We could use date pipe to convert NgbDateStruct to string but for reverse conversion I didn't find easy solution with build in tools.

PR #256, opened 2017-11-23
JR2-508 Implement UI for job search counter pill

Invizible:
Not finished, template will be updated once mockups are provided

### DIFF
--- a/src/main/webapp/app/home/tools/job-search-tool/job-search-tool.component.html
+++ b/src/main/webapp/app/home/tools/job-search-tool/job-search-tool.component.html
@@ -1,26 +1,36 @@
-<div>
-    <form [formGroup]="jobSearchForm" novalidate (submit)="search(jobSearchForm.value)">
-        <div class="form-group form-group--wide">
-            <jr2-typeahead-multiselect
-                formControlName="baseQuery"
-                [itemLoader]="fetchOccupationSuggestions"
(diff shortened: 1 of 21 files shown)


src/main/webapp/app/home/tools/job-search-tool/job-search-tool.component.html
```diff
@@ -1,26 +1,36 @@
...
+        <form id="ngForm" #jobForm="ngForm" [formGroup]="jobSearchForm" novalidate (submit)="search(jobSearchForm.value)">
+            <div class="form-group form-group--wide">
+                <jr2-typeahead-multiselect
+                    formControlName="baseQuery"
+                    [itemLoader]="fetchOccupationSuggestions"
+                    [placeHolder]="'job-search.toolbar.occupation.placeholder' | translate"></jr2-typeahead-multiselect>
 
+            </div>
+            <div class="form-group form-group--wide form-group--icon">
+                <label class="sr-only">Ort, Region, Kanton</label>
```
varusakov: I think we need to use translate pipe here

birom: We use the label on an inconsistent way. Sometimes we have a label with sr-only class and sometimes there is not any label. I'd suggest to remove the label and review this page after we define a strategy regarding accessibility.

PR #258, opened 2017-11-24
JR2-519 Remove invalid language skills in candidate search request

varusakov:


src/main/webapp/app/candidate-search/state-management/reducers/candidate-search.reducers.ts
```diff
@@ -9,6 +13,14 @@ import {
     SHOW_CANDIDATE_LIST_ERROR
 } from '../actions/candidate-search.actions';
 
+function removeInvalidLanguageSkills(filter: CandidateSearchFilter): CandidateSearchFilter {
```
birom: I think the search-request-mapper.ts is responsible for creating the proper request. 
Pease move the cleanup logic there.

varusakov: In that case we need also clean up language skills when encoding CandidateSearchFilter to base64 string. From my POV it's good to perform clean up in one place.

birom: If we do not want to allow a language without a value (see screenshot below), we should change the language input control or add some validator.

image

If it is a valid UI state, we should create a proper request when we call the REST endpoint.

I think in this case there is no need to be restrictive, because it is a filter filter and not a data input form.

varusakov: Fixed

PR #265, opened 2017-11-29
JR2-201 Add phone and organization fields for User management

Invizible:


src/main/java/ch/admin/seco/jobroom/service/dto/UserDTO.java
```diff
@@ -39,6 +39,8 @@
     @Size(min = 5, max = 100)
     private String email;
 
+    private String phone;
```
varusakov: Should we add validation constraints to phone field?

src/main/webapp/app/admin/user-management/user-management-dialog.component.ts
```diff
@@ -76,11 +104,13 @@ export class UserDialogComponent implements OnInit, OnDestroy {
 
     ngOnInit() {
         this.routeSub = this.route.params.subscribe((params) => {
-            if ( params['login'] ) {
-                this.modalRef = this.userModalService.open(UserMgmtDialogComponent as Component, params['login']);
-            } else {
-                this.modalRef = this.userModalService.open(UserMgmtDialogComponent as Component);
-            }
+            setTimeout(() => {
```
birom: Why do we need the setTimeout()?

Invizible: Because of this error: https://github.com/angular/angular/issues/17572

birom: Good to know. Thank you!

src/main/java/ch/admin/seco/jobroom/service/dto/UserDTO.java
Invizible: added

PR #269, opened 2017-11-30
Feature/jr2 525 url for stakeholders

Invizible:


src/main/webapp/app/home/state-management/effects/home.effects.ts
```diff
@@ -61,7 +67,38 @@ export class HomeEffects {
             }
         );
 
-    constructor(private actions$: Actions, private router: Router, private candidateService: CandidateService,
+    @Effect({ dispatch: false })
+    selectRoute$: Observable<Action> = this.actions$
+        .ofType(ROUTER_NAVIGATION)
```
birom: I created a separate effect class for the router ralated effect (router.effects.ts) in the candidate-search module.
We could also do this, to be consistent. Otherwise the solution is OK for me.

PR #272, opened 2017-12-04
JR3-522: Use OccupationLabelResource instead of OccupationResource

birom:
This needs the following PRs to be merged and deployed:
- https://github.com/alv-ch/job-service/pull/46
- https://github.com/alv-ch/jobroom-integration/pull/15


src/main/webapp/app/shared/reference-service/occupation-label.service.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * This service is supposed to be a data service. The UI components should use the OccupationPresentationService
+ * instead of this.
+ */
+@Injectable()
+export class OccupationLabelService {
+    constructor(private http: Http) {
+    }
+
+    suggestOccupation(prefix: string, types = ['avam', 'bfs', 'x28']): Observable<OccupationLabelAutocomplete> {
```
michael-wirth: the result set for the job search should contain "avam", "x28", "sbn3" and "snb5" label/codes -- same for the candidate search (as far there was no other agreement with tony)

if neither "sbn3" or "sbn5" is defined then the referenceservice should not return any classification values (this is a bug and I already opened a ticket for that)

but the the job publication we will only require the "avam" code

birom: I moved the default codes to the OccupationPresentationService and did the following modifications:
- I fixed the parameters for the job search.
- According to Toni the candidate search should only contain avam codes

PR #279, opened 2017-12-08
Feature/ setup ci jobpublicationservice

pellenberger:


devops.gradle
```diff
@@ -25,6 +25,9 @@ task copyArtifacts {
         exec {
             commandLine 'mvn', '-q', 'dependency:get', '-DrepoUrl=https://alvch.jfrog.io/alvch/libs-snapshots-local/', '-Dartifact=ch.admin.seco.jobroom.integration:jobroom_batch:LATEST:jar', '-Dtransitive=false', "-Ddest=${buildDir}/docker/dockerfiles/jobroom_batch/jobroom_batch.jar"
         }
+        exec {
+            commandLine 'mvn', '-q', 'dependency:get', '-DrepoUrl=https://alvch.jfrog.io/alvch/libs-snapshots-local/', '-Dartifact=ch.admin.seco.service.jobpublication:jobpublicationservice:LATEST:war', '-Dtransitive=false', "-Ddest=${buildDir}/docker/dockerfiles/jobpublicationservice/jobpublicationservice.war"
```
michael-wirth: shouldn't this be a jar? does the archive contain some web resources? i guest this is because of older jhipster version 4.5.6 and Spring Boot 1. As this project doesn't require elasticsearch - there is no need to use Spring Boot 2 already

pellenberger: I changed and generate now an executable jar.

PR #290, opened 2017-12-13
Feature/jr2 546 job publication detail view

varusakov:
Print functionality in progress


src/main/webapp/app/dashboard/dashboard-routing.module.ts
```diff
@@ -23,6 +25,15 @@ const routes = [
         component: DashboardComponent,
         canActivate: [PEAGuard],
         data: { pageTitle: 'dashboard.title' }
+    },
+    {
+        path: 'job-publication-detail/:id',
+        component: JobPublicationDetailComponent,
+        canActivate: [PEAGuard],
```
pellenberger: This page should also be accessible for anonymous users. 
It should be possible to display the job publication detail page without being logged in and using a link like http://dev.job-room.ch/job-publication-detail/:id?accessToken=:token

birom: Who generates the accessToken token and how should we check it? Could you please add this to the jira issue?

PR #294, opened 2017-12-15
JR2-468 Job publication edit view component

Invizible:


src/main/webapp/app/shared/job-publication/job-publication.model.ts
```diff
@@ -98,17 +106,25 @@ export enum Status {
     UNKNOWN
 }
 
+export enum Locale {
+    DE,
+    FR,
+    IT,
+    EN
```
pellenberger: **EN** is not a valid Locale to send to the API. If the locale is set to **en** in the frontend, send **DE** to the API.

src/main/webapp/i18n/de/home.json
```diff
@@ -101,7 +101,32 @@
                 },
                 "reset": "Zur\u00fccksetzen",
                 "submit": "Job publizieren",
-                "workload": "Arbeitspensum"
+                "submit.success": "Job publication saved!",
+                "submit.error": "Job publication could not be saved!",
+                "workload": "Arbeitspensum",
+                "education-level": {
+                    "PRIMAR_OBLIGATORISCHE_SCHULE": "PRIMAR OBLIGATORISCHE SCHULE",
```
pellenberger: I sent you the translations in DE, FR and IT

pellenberger: Same for education levels.

Invizible: Ok, will update

src/main/webapp/app/shared/job-publication/job-publication.model.ts
Invizible: will remove

PR #303, opened 2017-12-18
JR2-546 Add info message on job publication cancellation

varusakov:


src/main/webapp/app/dashboard/job-publication-detail/job-publication-detail.component.ts
```diff
@@ -1,35 +1,64 @@
...
-                private jobPublicationService: JobPublicationService) {
+                private jobPublicationService: JobPublicationService,
+                private store: Store<JobPublicationCancelState>,
+                private updates$: Actions) {
     }
 
     ngOnInit() {
         this.jobPublication$ = this.route.data
             .map((data) => data['jobPublication']);
+        this.subscribeOnEffect(JOB_PUBLICATION_CANCELED, () => this.cancelledSuccessfully = true);
```
birom: The cancelledSuccessfully and cancelledWithError flags should also be part of the state.
We should create a selector for these instead of subscribing to the Actions observable.

src/main/webapp/app/dashboard/state-management/state/job-publication-cancel.state.ts
```diff
@@ -0,0 +1,33 @@
...
+        jobCenter: boolean;
+        privateAgency: boolean;
+        self: boolean
+    }
+}
+
+export interface JobPublicationCancelState {
+    id?: string;
+    accessToken?: string;
+    modalRef?: NgbModalRef;
```
birom: The modalRef sould not be part of the state. We have to find a better solution to close the dialog.

birom:
I open an other PR because of my changes.

PR #308, opened 2017-12-19
JR2-552: Fix language selector labels

birom:
-add label to the skills
-preselect the lowest possible oral and written skill


src/main/webapp/app/home/tools/job-publication-tool/language-skills/language-skills.component.html
```diff
@@ -41,9 +40,8 @@
             </div>
             <div class="col">
                 <div class="form-group">
-                    <label class="sr-only" [for]="'written-' + i" translate="global.reference.language.written"></label>
+                    <label class="small" [for]="'written-' + i" translate="global.reference.language.written"></label>
                     <select [id]="'written-'+i" class="form-control custom-select" formControlName="written">
-                        <option [ngValue]="null">{{'global.reference.language.written'| translate}}</option>
                         <option *ngFor="let languageLevel of languageLevels | keys" [ngValue]="languageLevel.value">
```
birom: Are we use intentionally the string value instead of he enum key? I also saw this at other places eg.: educationLevels.

PR #309, opened 2017-12-19
JR2-484 Add size limits for job publication text fields

Invizible:


src/main/webapp/app/shared/validation/text-size-limit.directive.ts
```diff
@@ -0,0 +1,42 @@
+import { Directive, ElementRef, HostListener, Input, OnDestroy, OnInit } from '@angular/core';
+import { NgControl } from '@angular/forms';
+import { Subject } from 'rxjs/Subject';
+
+@Directive({
+    selector: '[jhiTextSizeLimit]'
```
birom: Our components are prefixed with _jr2_.

PR #319, opened 2017-12-21
Feature/jr2 569 users search

Invizible:


src/main/java/ch/admin/seco/jobroom/service/OrganizationSuggestionService.java
```diff
@@ -39,11 +40,13 @@ public OrganizationAutocompleteDTO suggest(String prefix, int resultSize) {
 
         SearchResponse suggest = elasticsearchTemplate.suggest(suggestBuilder, Organization.class);
 
-        List<OrganizationSuggestionDTO> suggestions = suggest.getSuggest()
-            .<CompletionSuggestion>getSuggestion("organizations").getEntries().stream()
-            .flatMap(item -> item.getOptions().stream())
-            .map(this::convertToOrganizationSuggestion)
-            .collect(Collectors.toList());
+        List<OrganizationSuggestionDTO> suggestions = suggest.getSuggest() != null
```
michael-wirth: just use a simple `null` guard
avoid long tertiary operators

```
if (Objects.isNull(suggest.getSuggest()) {
  return Collections.emptyList();
}
```

PR #320, opened 2017-12-21
Kof logging

HomoCodens:
Adds logging of various site-usage events.

I'm not in the office next week so responses will be slow.

### DIFF
--- a/src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.html
+++ b/src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.html
@@ -24,7 +24,7 @@
                             {{'candidate-detail.job-center.show-contact' | translate}}
                         </button>
 
-                        <p *ngIf="RAVContactVisible">
+                        <p *ngIf="RAVContactVisible == true">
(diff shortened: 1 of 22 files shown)


HomoCodens:
> the "logging" should be isolated from the functional code and follow the Separation of Concerns design principle. I still don't know what this requirement is about.

I admit I stuck much of the code in the first place that seemed to work. The thing is I'm actually only supposed to work on this 20% of my time and time is pressing since it is very important for our project that there be no interruption in the data. 

> As we use ngrx we could use some event driven hooks for the logging... We should sit together and define a concept/design how to do the logging

Where exactly does the logging violate Separation of Concerns? Only in the reducers/effects or also in the candidate detail component?
Could you go into detail a little or comment the code in the relevant places? Or are more fundamental changes required?

If you want to talk I'm on HipChat or we could also arrange for a meeting if you like.

birom:
The user tracking logic has been already implemented in a [ngrx effect](https://github.com/alv-ch/jobroom2/blob/develop/src/main/webapp/app/shared/state-management/effects/user-tracking.effects.ts).

PR #333, opened 2017-12-27
Bugfix/jr2 622 set expiration date

michael-wirth:


michael-wirth:
the build fails because of a bug in the JR2-603, which will be solved when the JR2-603 is merged

src/main/webapp/app/candidate-search/services/candidate.service.ts
```diff
@@ -82,8 +82,11 @@ export class CandidateService {
             .filter((jobExperience) => jobExperience.wanted);
 
         if (occupationCode) {
+            const parts = occupationCode.split(':');
+            const code = parts[parts.length - 1];
             const jobExperience = jobExperiences
-                .find((_jobExperience) => String(_jobExperience.occupationCode) === String(occupationCode));
+                .find((_jobExperience) => String(_jobExperience.occupationCode) === String(code));
+            console.log(occupationCode, jobExperience);
```
varusakov: Could you remove this?

michael-wirth: done

PR #337, opened 2017-12-27
JR2-609 Populate job publication form with user data

Invizible:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -212,10 +214,10 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
             },
```
michael-wirth: additional requirement:

The company should also be populated with the organization data

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -212,10 +214,10 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
...
             contact: {
                 salutation: null,
-                firstName: '',
-                lastName: '',
-                phoneNumber: '',
-                email: ''
+                firstName: this.principal.firstName,
```
michael-wirth: is it checked that `this.principal` is never null? What happens when the user is not logged in?

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
michael-wirth: the company is in another jira ticket

PR #338, opened 2017-12-28
JR2-429 Clean up webpack configs

varusakov:


webpack/webpack.dev.js
```diff
@@ -9,96 +9,123 @@ const utils = require('./utils.js');
...
-            ],
-            target: 'http://127.0.0.1:8080',
+const getProxies = (enviroment) => {
+    const target = enviroment && enviroment.target ? enviroment.target : 'http://127.0.0.1:8080';
+    if (enviroment && enviroment.localUrl && enviroment.localPort) {
+        const localUrl = enviroment.localUrl;
+        const devUrls = URLS.filter((url) => url !== localUrl);
+        return [{
+            context: devUrls,
+            target: 'https://dev.job-room.ch:8443',
```
michael-wirth: is there a special reason to use `https` here? whe should be consistent and either use http or https. if there are no restrictions i'd use `http`

varusakov: You are right, thanks

PR #339, opened 2017-12-28
JR2-616 Reindex Organization

michael-wirth:


src/main/java/ch/admin/seco/jobroom/service/search/UserSearchService.java
```diff
@@ -34,19 +31,17 @@ public UserSearchService(UserSearchRepository userSearchRepository, UserDocument
 
     public Page<UserDTO> searchByQuery(String query, Pageable pageable) {
         QueryStringQueryBuilder queryBuilder = queryStringQuery(query)
+            .defaultOperator(Operator.AND)
```
michael-wirth: the AND operator is better to find a specific user  as 
- searching for firstname and lastname 
- searching for email. the analyzer will splitt up the email in multiple terms with delimiter [@.-]

PR #342, opened 2017-12-29
JR2-635 Reset candidate search after language change

varusakov:
Resetting of job search and homepage tools will be done in separate PR


src/main/webapp/app/candidate-search/candidate-search-filter/candidate-search-filter.component.ts
```diff
@@ -74,6 +68,13 @@ export class CandidateSearchFilterComponent implements OnInit, OnDestroy {
         this.candidateSearchUrl$ = this.store.select(getSearchFilter)
             .map((filter: CandidateSearchFilter) =>
                 `${window.location.href}/?searchFilter=${this.candidateService.encodeURISearchFilter(filter)}`);
+        this.store.select(getResetSearchFilter)
```
birom: The language change must be handled general. The landing page forms and the job search also have to be reset.
Please write todo here to fix this.

src/main/webapp/app/candidate-search/candidate-search.component.ts
```diff
@@ -56,6 +55,17 @@ export class CandidateSearchComponent {
             .map(([filter, options]) => residenceMapper(filter, options));
     }
 
+    ngOnInit(): void {
+        this.languageChangeSubscription = this.translateService.onLangChange
+            .subscribe((_: LangChangeEvent) => {
+                this.store.dispatch(new ResetSearchFilterAction())
```
birom: I think we sould handle this more general, because we need this on the landing page forms and the job search too.
Could you please write here a todo for the https://alv-ch.atlassian.net/browse/JR2-636.

PR #343, opened 2018-01-02
Release/rc2

michael-wirth:


src/main/resources/config/application-dev.yml
```diff
@@ -19,6 +19,7 @@ eureka:
     client:
         service-url:
             defaultZone: http://admin:${jhipster.registry.password}@localhost:8761/eureka/
+        enabled: false
```
varusakov: Why this setting was disabled?

michael-wirth: it was disabled for local testing. my fault. will revert it in the next commit

PR #345, opened 2018-01-03
JR2-636 Reset job search after language change

varusakov:


birom:
This PR implements the same: https://github.com/alv-ch/jobroom2/pull/350 feature.

PR #352, opened 2018-01-04
JR2-592: Use share() to avoid multiple requests

birom:


birom:
I think in the future we should consider to use ngrx store for the reference data too.

PR #364, opened 2018-01-10
Fix #JR2-588

tdascoli:


src/main/webapp/i18n/de/global.json
```diff
@@ -55,7 +55,8 @@
                 "text": "Abk\u00fcrzungen"
             },
             "legal": {
-                "text": "Rechtliche Grundlagen"
+                "text": "Rechtliche Hinweise",
+                "url": "https://www.arbeit.swiss/home/service/rechtliche-hinweise.html"
```
michael-wirth: It would be better to put external URLs into the `application.yml` configuration files instead of putting them into the translation files, there might be more language depending URLs and Links

for now we can go with this solution, but we need to keep track of this technical dept

PR #366, opened 2018-01-10
JR2-602 Fix language change event

michael-wirth:
- first change the language cookie NG_TRANSLATE_LANG_KEY and
  then call all the event listener
- also change the 'last' job title
- use language cookie instead of 'language' param


src/main/webapp/app/shared/reference-service/occupation-label.service.ts
```diff
@@ -79,23 +72,17 @@ export class OccupationLabelService {
...
-        return this.http.get(OCCUPATION_LABEL_RESOURCE_URL, options)
+        return this.http.get(OCCUPATION_LABEL_RESOURCE_URL)
             .map((res: Response) => <OccupationLabelData>res.json())
     }
 
     getOccupationLabelsByKey(key: string): Observable<OccupationLabelData> {
-        const params: URLSearchParams = new URLSearchParams();
-        params.set('lang', this.translateService.currentLang);
-        return this.http.get(`${OCCUPATION_LABEL_RESOURCE_URL}/${key.replace(':', '/')}`, { params })
+        return this.http.get(`${OCCUPATION_LABEL_RESOURCE_URL}/${key.replace(':', '/')}`)
```
varusakov: I added lang param to prevent browser caching. Does it work without it?

michael-wirth: yes it works great - 

btw the param name is `language` and not just 'lang'

varusakov: ok, thanks

PR #369, opened 2018-01-10
JR2-673 Share JobExperience list between several subscriptions

varusakov:


src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.ts
```diff
@@ -82,7 +82,7 @@ export class CandidateDetailComponent implements OnInit {
                     .map((jobExperiences) => jobExperiences.map(this.formatOccupationLabel(profile.gender)))
             })
             .map((experiences) => experiences.sort((a, b) => +b.lastJob - +a.lastJob))
-            .share();
+            .shareReplay();
```
birom: What was the problem with the share?

varusakov: `share` doesn't repeat events that happened before subscription is started. You can find short reference for this operators [here](https://medium.com/@_achou/rxswift-share-vs-replay-vs-sharereplay-bea99ac42168)

birom: Thank you!

PR #376, opened 2018-01-11
Sepearet Street/house number and add label

tdascoli:


michael-wirth:
hope will all these translation changes we won't have too many merge conflicts

PR #377, opened 2018-01-11
JR2-680 Add cache interceptor

michael-wirth:


src/main/webapp/app/blocks/interceptor/cache.interceptor.ts
```diff
@@ -0,0 +1,33 @@
...
+    requestIntercept(options?: RequestOptionsArgs): RequestOptionsArgs {
+        const requestOptions = new RequestOptions(options);
+        if (!options) {
+            options = new BaseRequestOptions();
+        }
+        if (!requestOptions.params) {
+            requestOptions.params = new URLSearchParams();
+        }
+
+        requestOptions.params.set('_ng', this.base64.encode(this.cookieService.get('NG_TRANSLATE_LANG_KEY')));
```
michael-wirth: Do you have any good suggestion for the parameter name? It should be obfuscated a little bit

PR #391, opened 2018-01-12
JR2-692 Change query French and Italian 'query' translations

michael-wirth:


src/main/webapp/i18n/de/candidate-search.json
```diff
@@ -8,22 +8,22 @@
...
-                },
-                "one": {
-                    "with-locality": "1 {{query}} Kandidat in {{locality}}",
-                    "without-locality": "1 {{query}} Kandidat"
-                },
-                "other": {
-                    "with-locality": "{{count}} {{query}} Kandidaten in {{locality}}",
-                    "without-locality": "{{count}} {{query}} Kandidaten"
-                }
+                "many": "Mehr als {{count}} Kandidaten",
```
birom: I think we either need keys with "many" or with "other". In the code only "other" is used, but I think "many" would be better to be more consistent.

michael-wirth: agree, but we have to decide that with toni and thomas

birom: I have removed the unused keys, and changed the "other" part of the keys to "many".

PR #410, opened 2018-01-16
JR2-706 Move labels inline

varusakov:


src/main/webapp/app/shared/input-components/range-input/range-input.component.scss
```diff
@@ -0,0 +1,4 @@
+.range-select {
+    width: 50%;
```
birom: I think this could be 75%, because it looks now a little bit weird to me.
**With 50%**
selection_024

**With 75%**
selection_025

@tdascoli: Could you please have a look at it. Thank you!

varusakov: @birom It's not enough space with width 75% on candidate/job search pages.

michael-wirth: inline labels are awful and break the design - we should at least ask our web designer and UX guy Ingo

michael-wirth: @tdascoli I'd rather put the `30 Tagen` right next to `Online seit` - this looks draws apart

tonim42: The 75% version looks fine with me. The line break in "30 Tagen" should be fixed with JR2-703 but it appears it 's not fixed.

tonim42: @michael-wirth The current labels break the reading flow. If space is an issue, perhaps the best solution is to drop the 'von'.

varusakov: New version looks following:

**Job search**

screen shot 2018-01-16 at 15 58 53

**Job publication**

screen shot 2018-01-16 at 16 00 51

PR #417, opened 2018-01-16
JR2-687 Show error message on form submit

Invizible:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.html
```diff
@@ -491,7 +491,7 @@
                             {{ 'home.tools.job-publication.disclaimer' | translate }}
                         </label>
                         <small class="form-text text-danger"
-                               *ngIf="jobPublicationForm.get('disclaimer').invalid"
+                               *ngIf="form.submitted && jobPublicationForm.get('disclaimer').invalid"
```
birom: Do we need to hide the message on reset? If yes we could set type="reset" to the reset button,
otherwise the form.submitted flag remains true.

PR #423, opened 2018-01-18
Feature/jr2 664 gatling tests

varusakov:


src/test/gatling/user-files/simulations/JobroomGatlingTest.scala
```diff
@@ -0,0 +1,280 @@
...
+            .body(StringBody("""{"username":"manpower63", "password":"passwort01"}""")).asJSON
+            .check(header.get("Authorization").saveAs("agent_token"))
+    }
+
+    def loadJobPublications() = {
+        val index = scala.util.Random.nextInt(20)
+        http("Search Job Publications")
+            .post("/jobpublicationservice/api/_search/jobPublications")
+            .headers(headers_http_authenticated)
+            .body(StringBody("""{"email":"bruno.goncalves@manpower.ch"}""")).asJSON
```
birom: We need to create a dedicated test user and this email should not be in the source code.

PR #425, opened 2018-01-18
JR2-706 Fix layout for workload field

Invizible:


birom:
I looks good on Edge, Firefox, Chrome and mobile Chrome on Android 5.1.

varusakov:
How it looks in safari and chrome when width of the browser < 100%
screen shot 2018-01-18 at 16 18 38

PR #427, opened 2018-01-18
JR2-205 Change job publication detail application layout

michael-wirth:
image


src/main/webapp/app/dashboard/job-publication-detail/job-publication-detail.component.html
```diff
@@ -1,22 +1,25 @@
-<div class="container">
+<div class="container" xmlns="http://www.w3.org/1999/html">
```
birom: Why do we need the name space?

michael-wirth: i'll remove that...

I guess Intellij thinks I need it there - annoying

michael-wirth:
merged with JR2-716

PR #440, opened 2018-01-24
JR2-743 Fix work.swiss etc. links

tonim42:


birom:
Hi, I've just updated the json files.

PR #443, opened 2018-01-25
JR2-735 Lookup job by 'fingerprint' and open job details page

michael-wirth:
we need to define a proper job-not-found-message.

Is Juelien not yet registered here?


src/main/webapp/app/job-search/job-fingerprint-redirect/job-fingerprint-redirect.component.html
```diff
@@ -0,0 +1,4 @@
+<div class="container mt-md-5">
+    <h1>Sorry, this job is gone</h1>
```
birom: We also need a translation for this, or not?

PR #445, opened 2018-01-29
Bugfix/jr2 737 index organization name and external

varusakov:


src/main/java/ch/admin/seco/jobroom/service/dto/UserDTO.java
```diff
@@ -82,7 +83,10 @@ public UserDTO(User user) {
             user.getAuthorities().stream().map(Authority::getName)
                 .collect(Collectors.toSet())
         );
-        this.organizationId = getOrganizationId(user);
+        Optional.ofNullable(user.getOrganization()).ifPresent(organization -> {
```
birom: Please keep it simple and just use an if statement, thx.

michael-wirth: I agree, Optional is not designed to be used this way
Optional should be used only as return values. there is a whole blog on how to use it .i have to look it up

varusakov: Thanks, I simplified it

PR #447, opened 2018-01-30
Feature/jr2 655 filter is reset on language change both in job search and candidate search

birom:


src/main/webapp/app/candidate-search/state-management/effects/candidate-search.effects.ts
```diff
@@ -37,6 +39,8 @@ import {
 import { Router } from '@angular/router';
 import { CandidateProfile } from '../../services/candidate';
 import { WINDOW } from '../../../shared/shared-libs.module';
+import { RESET } from '../../../shared/state-management/actions/core.actions';
```
varusakov: Maybe it's better to rename `RESET` action to `LANGUAGE_CHANGE` or something similar?

varusakov: Maybe it's better to rename RESET action to LANGUAGE_CHANGE or something similar?

birom: Fixed, thx.

PR #448, opened 2018-01-31
Fix #JR2-757 add Link to contact form

tdascoli:


src/main/webapp/i18n/de/home.json
```diff
@@ -9,6 +9,10 @@
             "recruitmentAgencies": "Arbeitsvermittler",
             "title": "Job-Room bietet Leistungen f\u00fcr"
         },
+        "contactform": {
```
birom: Please also commit the translations.csv file.

birom: Please also commit the translations.csv file.

tdascoli: will fix that

PR #460, opened 2018-02-07
Feature/jr2 421 add reset button on search forms

varusakov:


birom:
We need translations for the candidate-search.toolbar.reset and job-search.toolbar.reset keys.

PR #472, opened 2018-02-14
Feature/jr2 766 Calendar work with keyboard

Invizible:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -61,6 +63,12 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
     @Input()
     userData: UserData;
 
+    @ViewChild('publicationStartDateEl')
+    publicationStartDateEl: ElementRef;
```
michael-wirth: what does the suffix `El` mean? is it for "ElementRef" maybe we should use not abbreviate the term to make it more clear. @birom what do you think? Do we have a naming convention for ElementRef? fields?

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -82,6 +90,23 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
 
     private unsubscribe$ = new Subject<void>();
 
+    private static ngbDateStructValidator(control: AbstractControl): ValidationErrors | null {
```
michael-wirth: what is this validator doing? what means `ngb`? Can we find a better name?

michael-wirth: Also describe the what the Validator is doing. Also provide valid and invalid formats as example comment

Invizible: I figured, we do not need this validator, will remove

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
birom: No, we do not really have.

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.html
```diff
@@ -151,6 +152,10 @@
                                                    aria-hidden="true"></i>
                                             </button>
                                         </div>
+                                        <jr2-input-error-message [formGroup]="jobPublicationForm" path="job.publicationStartDate.date"
+                                                                 [errorCodes]="['required', 'ngbDate.invalid']"
```
birom: Error code 'ngbDate.requiredBefore' is missing here.

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.html
```diff
@@ -182,20 +187,25 @@
...
+                                                   [minDate]="publicationEndDateMin"
+                                                   (focus)="endDatePicker.close()"/>
                                             <button class="input-group-addon"
                                                     (click)="endDatePicker.toggle(); startDatePicker.close()"
                                                     type="button"
                                                     [disabled]="publicationEndDateIsPermanent">
                                                 <i class="fa fa-calendar-o" aria-hidden="true"></i>
                                             </button>
                                         </div>
+                                        <jr2-input-error-message [formGroup]="jobPublicationForm" path="job.publicationEndDate.date"
```
birom: Error code 'ngbDate.requiredBefore' is missing here.

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
michael-wirth: @Invizible please rename the suffix `El` to `ElementRef` - it's longer but I think its clearer

PR #473, opened 2018-02-14
JR2-770 Organization search by name and city

varusakov:


src/main/java/ch/admin/seco/jobroom/service/OrganizationSuggestionService.java
```diff
@@ -2,74 +2,65 @@
...
     }
 
-    private SuggestBuilder buildSuggestBuilder(String prefix, int resultSize) {
-        CompletionSuggestionBuilder nameSuggestion = SuggestBuilders.completionSuggestion("suggestions")
-            .prefix(prefix)
-            .size(resultSize);
-        return new SuggestBuilder()
-            .addSuggestion("organizations", nameSuggestion);
+    private SearchQuery buildSearchQuery(String query, int resultSize) {
+        final MultiMatchQueryBuilder multiMatchQueryBuilder = new MultiMatchQueryBuilder(query, "name", "city")
```
michael-wirth: also include zipcode

michael-wirth: can we set the `searchAnalyzer` here? 

issue: search by `Pfaeffikon` doesn't match `Pfäffikon`

varusakov: @michael-wirth it's possible to set analyzer but it still not matching `Pfaeffikon`  `Pfäffikon`. I am investigating this issue now

PR #477, opened 2018-02-15
JR2-786 Add Tooltip for "Multivalue" autocomple component

Invizible:


src/main/webapp/app/shared/input-components/multiselect/multiselect.component.html
```diff
@@ -1,4 +1,5 @@
-<div class="input input-multiselect d-flex flex-row flex-wrap" (click)="focusInput()">
+<div class="input input-multiselect d-flex flex-row flex-wrap" (click)="focusInput()"
+     placement="bottom" [ngbTooltip]="tooltip">
```
birom: Should not we close the tooltip, as the user starts typing? I think it is disrupting if the tooltip stays opened.

Invizible: agree. also, do we need to show tooltip on focus instead of hover?

birom: For me it is OK if we show the tooltip on focus, but the PO takes the final decison.

src/main/webapp/app/shared/input-components/multiselect/multiselect.component.html
```diff
@@ -1,12 +1,19 @@
...
+     placement="bottom" [ngbTooltip]="tooltip" triggers="manual"
+     #t="ngbTooltip" (mouseenter)="t.open()" (mouseleave)="t.close()">
     <canvas #canvas height="0" width="0"></canvas>
     <span class="badge input-multiselect__tag" *ngFor="let item of selectedItems">
     {{item}}<span class="fa fa-times remove-button" aria-hidden="true" (click)="removeItem(item)"></span>
     </span>
 
     <input [id]="id" #input class="input input-multiselect__input" autofocus name="inputValue"
-           (keydown)="handleKeyDown($event)"
+           (keydown)="handleKeyDown($event); t.close()"
```
Invizible: So I've added additional triggers: tooltip will work on hover as before and additionally will show up on focus, and will disappear when we start typing

PR #493, opened 2018-02-24
Fix sort request

michael-wirth:


michael-wirth:
PLease check all 'new HttpParams()' handling as is will now not update the current instance but create a new one

PR #500, opened 2018-03-01
Fix bad-taste party in candidate-search filter

tdascoli:


src/main/webapp/app/candidate-search/candidate-search-toolbar/candidate-search-toolbar.component.html
```diff
@@ -12,14 +12,15 @@
                     [placeHolder]="'candidate-search.occupations.placeholder' | translate"></jr2-typeahead-multiselect>
             </div>
 
-            <div class="form-group form-group--wide w-100">
+            <div class="form-group form-group--wide">
```
birom: I think it breaks the layout on mobile:

selection_042

tdascoli: you're right... back to the planning board :)

tdascoli: @birom please verify again

PR #515, opened 2018-03-21
JR2-862 Populate user language for job registration

michael-wirth:
quick fix for the job registration problem for loged-in users


src/main/webapp/app/shared/job-publication/job-publication.service.ts
```diff
@@ -27,7 +27,7 @@ export class JobPublicationService {
 
     private getJobPublicationLocale(): Observable<string> {
         return this.principal.isAuthenticated()
-            ? Observable.fromPromise(this.principal.identity())
+            ? Observable.fromPromise(this.principal.identity().then((identity) => identity.langKey))
```
birom: I'd rather use the the Observable.map operator instead of this because we use most of the time 
Observables instead of Promises.
eg.:
```
 return this.principal.isAuthenticated()
            ? Observable.fromPromise(this.principal.identity())
                .map((identity) => identity.langKey.toUpperCase())
            : Observable.of(this.translateService.currentLang);
```

PR #531, opened 2018-04-30
JR2-887 Refactor job search

Invizible:


src/main/webapp/app/job-search/job-search-list-item/job-search-list-item.component.ts
```diff
@@ -1,10 +1,27 @@
...
+    @Input() job: JobAdvertisement;
+    jobDescription$: Observable<JobDescription>;
+
+    constructor(
+        private translateService: TranslateService
+    ) {
+    }
+
+    ngOnInit(): void {
+        this.jobDescription$ = Observable.merge(
```
birom: Please move the mapping into the JobSearchComponent.

PR #533, opened 2018-04-30
JR2-887 Fix publicContact section

Invizible:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-mapper.ts
```diff
@@ -218,13 +220,15 @@ export class JobPublicationMapper {
...
         };
 
-        jobAd.publicContact = {
-            salutation: <Salutation>Salutation[jobPublicationForm.publicContact.salutation],
-            firstName: jobPublicationForm.publicContact.firstName,
-            lastName: jobPublicationForm.publicContact.lastName,
-            phone: jobPublicationForm.publicContact.phoneNumber,
-            email: jobPublicationForm.publicContact.email
-        };
+        if (!JobPublicationMapper.allFieldsSet(jobPublicationForm.publicContact)) {
```
cjpsycos: What's happening if not all the fields are set? is an error message displayed to the user or we just pass nothing to the backend?

Invizible: once user types something to any field in publicContact section all fields become required, so yes, he will see errors

PR #548, opened 2018-05-14
JR2-781 Change elastic reindex url for job-ad service

Invizible:
Related to https://github.com/alv-ch/job-ad-service/pull/78/files

### DIFF
--- a/src/main/webapp/app/admin/elasticsearch-reindex/elasticsearch-reindex.service.ts
+++ b/src/main/webapp/app/admin/elasticsearch-reindex/elasticsearch-reindex.service.ts
@@ -7,7 +7,7 @@ export class ElasticsearchReindexService {
 
     private readonly documentUrls = {
         'users': 'api/elasticsearch/index',
-        'jobs': 'jobadservice/api/jobAdvertisements/elasticsearch/index',
+        'jobs': 'jobadservice/api/elasticsearch/index',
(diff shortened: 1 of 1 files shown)


src/main/webapp/app/admin/elasticsearch-reindex/elasticsearch-reindex.service.ts
```diff
@@ -7,7 +7,7 @@ export class ElasticsearchReindexService {
 
     private readonly documentUrls = {
         'users': 'api/elasticsearch/index',
-        'jobs': 'jobadservice/api/jobAdvertisements/elasticsearch/index',
+        'jobs': 'jobadservice/api/elasticsearch/index',
```
birom: Why do we need this? The job ad service reindex URL is: 
```/api/jobAdvertisements/elasticsearch/index```

Invizible: I moved old reindex method to the new controller which reindexes jobs and apiUsers

Invizible: https://github.com/alv-ch/job-ad-service/pull/78/files#diff-571cfcf12f1de71a6d3799a20379947fR18

PR #560, opened 2018-05-29
Bugfix/jr2 968 system message

Invizible:


src/main/webapp/app/home/system-notification/system.notification.component.html
```diff
@@ -1,4 +1,8 @@
 <div class="system-notification" *ngFor="let activeSystemNotification of activeSystemNotifications">
-    <span class="warning-icon"><i class="fa fa-exclamation"></i></span>
-    <span class="system-notification-text">{{getCurrentLanguageCode(activeSystemNotification)}}</span>
+    <div class="container">
+        <span class="warning-icon"><i class="fa fa-exclamation"></i></span>
+        <span class="system-notification-text">{{getCurrentLanguageCode(activeSystemNotification)}}</span>
+        <a class="float-right" (click)="close(activeSystemNotification.id)">&times;</a>
+    </div>
```
fabiangotzen: From what i get out of the description, the issue here is the spacing between the text and the warning icon. So removing the container div and just adding a margin-left to the system-notification-text should solve it. You can see the original mockup in JR2-874 to clarify.

src/main/webapp/app/home/state-management/state/home.state.ts
```diff
@@ -23,6 +24,8 @@ export const getSystemNotificationsEntities = createSelector(getSystemNotificati
 export const getActiveSystemNotifications = createSelector(
     getSystemNotificationsEntities,
     (entities) => {
-        return Object.keys(entities).map((id) => entities[id]);
+        return Object.keys(entities)
+            .map((id) => entities[id])
+            .filter((entity: SystemNotification) => entity.active);
```
fabiangotzen: I'm not sure this is necessary because the filtering should already happen in the backend, see SystemNotificationRepository & Service. If there are issues with it I think it would be better to fix it there.

PR #561, opened 2018-05-30
JR2-792 Fix number formatting

varusakov:


src/main/webapp/app/home/tools/candidate-search-tool/candidate-search-tool.component.ts
```diff
@@ -78,7 +81,13 @@ export class CandidateSearchToolComponent implements OnInit, OnDestroy, OnChange
         this.store.dispatch(new CandidateSearchToolSubmittedAction(formValue));
     }
 
-    getBadgeKey() {
+    getBadgeText() {
+        return this.translateService.stream(this.getBadgeKey(), {
+            count: new LocaleAwareDecimalPipe(this.translateService).transform(this.candidateSearchToolModel.totalCount)
```
Invizible: You can inject pipe into component, no need to inject it's dependencies and create pipe manually

varusakov: Tnx, injecting of pipe is better approach

PR #565, opened 2018-06-04
[jr2-921] Registration roles selection page

wwebec:


.travis.yml
```diff
@@ -35,20 +35,26 @@ install:
 - yarn install
 script:
 - yarn test
-- ./gradlew check jacocoTestReport sonarqube -Pprod
+- ./gradlew check -x test jacocoTestReport -Pprod
```
birom: Why do we skip the tests?

wwebec:
Wrong destination branch, it should be merged with feature/eiam-integration-poc instead.

PR #566, opened 2018-06-04
Feature/jr2 921 registration roles selection page

wwebec:


src/main/webapp/app/account/registration/select/select.component.html
```diff
@@ -1,3 +1,118 @@
-<p>
-  select works!
-</p>
+<div class="registration-questionnaire">
+    <form [formGroup]="roleForm" (ngSubmit)="next()" novalidate>
+
+        <div class="registration-questionnaire-header">
```
birom: We use the [BEM](http://getbem.com/naming/) naming conventions in the [styleguide](https://github.com/alv-ch/styleguide) project and here as well. 
Can the _registration-questionnaire-header_  and the _registration-questionnaire-paragraph_ exist outside of a registration-questionnaire? If not please rename the classes to _registration-questionnaire__header_ and _registration-questionnaire__paragraph_ to follow the BEM guidlenies. Thank you!

wwebec:
PR is already merged into **feature/registration-integration** by @stefanfitzi

PR #575, opened 2018-06-06
JR2-995 Remove optional label for occupation field

Invizible:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.html
```diff
@@ -24,7 +24,6 @@
                         <div class="form-group">
                             <label>
                                 {{ 'home.tools.job-publication.occupation.label' | translate }}
-                                <span class="optional">{{ 'home.tools.job-publication.optional' | translate }}</span>
```
varusakov: Should we also remove 'home.tools.job-publication.optional' label from translations.csv?

Invizible: No, it's used in several places on this page

PR #614, opened 2018-06-18
JR2-1001 Add validation for job application channels

varusakov:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.html
```diff
@@ -616,6 +612,11 @@
                             </small>
                         </div>
                     </div>
+
+                    <jr2-input-error-message [formGroup]="jobPublicationForm" path="application"
```
birom: The display of the error message is buggy. We've already discussed it.

PR #631, opened 2018-06-21
Bugfix/jr2 1014 remaining translations

fabiangotzen:


src/main/webapp/i18n/de/registration-questionnaire.json
```diff
@@ -1,20 +1,25 @@
...
-            "existing": "Ich bin ein bestehender Job-Room Nutzer",
             "message": "Ich m\u00f6chte arbeit.swiss als <b>privater Arbeitsvermittler</b> nutzen",
             "new": "Ich melde mich zum ersten Mal bei arbeit.swiss bzw. job-room an",
-            "msg": "Ich m\u00f6chte arbeit.swiss als [<b>privater Arbeitsvermittler</b>] nutzen"
+            "existing": "Ich bin ein bestehender Job-Room Nutzer"
         },
         "termsOfUse": {
             "accept": "Ich akzeptiere die Nutzungsbedingungen",
-            "link": "Nutzungsbedingungen anzeigen"
+            "link": "<a href=\"/docs/Job-Room-AGB-20180530_de.pdf\" target=\"_blank\">Nutzungsbedingungen anzeigen</a>"
```
cjpsycos: Does this work?
previously it was the text of the link, now it's the link. has it also be changed on the page?

fabiangotzen: It redirects correctly, so yes it should work when there is a pdf ready.

PR #667, opened 2018-06-26
JR2-1098 Replace companyName with a companyId in PEA search request

Invizible:
https://github.com/alv-ch/job-ad-service/pull/120


src/main/webapp/app/dashboard/state-management/effects/pea-dashboard.effects.ts
```diff
@@ -74,7 +74,7 @@ export class PEADashboardEffects {
         const { jobTitle, onlineSinceDays } = filter;
         return {
             body: {
-                companyName: currentUser.companyName,
+                companyId: currentUser.companyId,
```
birom: It is O.K. for me, but we can merge this only with https://github.com/alv-ch/job-ad-service/pull/120 .

PR #705, opened 2018-07-12
JR2-1194 Return system notification id in create response

varusakov:


src/main/java/ch/admin/seco/jobroom/service/SystemNotificationService.java
```diff
@@ -16,70 +16,56 @@
...
 
 @Service
 @Transactional
 public class SystemNotificationService {
 
-    private SystemNotificationRepository systemNotificationRepository;
     private final Logger log = LoggerFactory.getLogger(SystemNotificationService.class);
 
-    public SystemNotificationService(SystemNotificationRepository systemNotificationRepository) {
+    private SystemNotificationRepository systemNotificationRepository;
```
birom: Why do we need all these changes? For me it seems that we are having some issues with the time zone handling.
Does the new mapper solve this?

varusakov: @birom Issue with time handling was on front end

birom: @varusakov If we do not need the other changes please revert them. Thank you!

src/main/webapp/app/shared/utils/date-utils.ts
```diff
@@ -49,4 +49,10 @@ export class DateUtils {
     static convertNgbDateStructToString(date: NgbDateStruct): string {
         return date ? moment(DateUtils.mapNgbDateStructToDate(date)).format('YYYY-MM-DD') : null;
     }
+
+    static convertNgbDateTimeToISOWithoutZone(date: NgbDateStruct, time: NgbTimeStruct): string {
+        return moment().utc().date(date.day).month(date.month - 1).year(date.year)
```
birom: Does it make any difference if we use ```moment().utc()``` or just simply ```moment()``` when we do the formatting with the ```YYYY-MM-DDTHH:mm:ss.sss``` pattern?

src/main/java/ch/admin/seco/jobroom/service/SystemNotificationService.java
varusakov: @birom we already have concept of mappers in jobroom. So I think it's good idea to move object mapping from service to them.

src/main/webapp/app/shared/utils/date-utils.ts
varusakov: Thank you, you are right. Simple `moment()` works perfectly

src/main/java/ch/admin/seco/jobroom/service/SystemNotificationService.java
birom: @cjpsycos What do you think?

cjpsycos: I hate this mapper things. It is less readable and usually change-errors are seen, when it's deployed. with manually mapping you see what is going where and e.g. type-changes are recognized instantly.

varusakov: Ok, I'll revert this changes

PR #717, opened 2018-08-01
JR2-1229 Fix api user search

varusakov:


src/main/webapp/app/admin/api-user-management/state-management/state/api-user-management.state.ts
```diff
@@ -17,7 +17,7 @@ export interface ApiUserManagementState {
 export const initialState: ApiUserManagementState = {
     filter: {
         query: null,
-        sort: 'apiUser.username,asc'
+        sort: 'apiUser.username.keyword,asc'
```
birom: Why do we need the keyword suffix? Do we need any change in the service?

PR #720, opened 2018-08-06
JR2-1176 Remove "keine Angabe" option

varusakov:


src/main/webapp/app/shared/reference-service/occupation-presentation.service.ts
```diff
@@ -141,11 +143,19 @@ export class OccupationPresentationService {
                 ? Observable.of([])
                 : this.occupationLabelService.suggestOccupation(prefix, ['avam'])
                     .map((autoComplete: OccupationLabelAutocomplete) => autoComplete.occupations)
-                    .map((occupations: OccupationLabelSuggestion[]) =>
-                        occupations.map((o: OccupationLabelSuggestion) => Object.assign({}, {
-                            key: new OccupationCode(o.code, 'avam').toString(),
-                            label: o.label
-                        })))
+                    .map((occupations: OccupationLabelSuggestion[]) => {
```
birom: According to @tonim42 the "keine Angabe" option must be filtered only on the job publication form (We should keep it at other places). For this reason please do the filtering in the ``fetchJobPublicationOccupationSuggestions`` method insead of this.

varusakov: @birom thank you for your input. I've moved logic to JobPublicationToolComponent.

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -113,7 +114,8 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
     }
 
     ngOnInit(): void {
-        this.fetchOccupationSuggestions = this.occupationPresentationService.fetchJobPublicationOccupationSuggestions;
+        this.fetchOccupationSuggestions = (prefix$: Observable<string>) => this.occupationPresentationService.fetchJobPublicationOccupationSuggestions(prefix$)
```
birom: Hi @varusakov, why did not you change the ```OccupationPresentationService.fetchJobPublicationOccupationSuggestions```? The component could stay simple as it is without such a logic and a service could have a test.

PR #723, opened 2018-08-10
JR2-1261 Adjust job-ad details page layout

varusakov:


tonim42:
Layout improvement:

image

src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -129,11 +129,13 @@ <h2 class="text--light" [innerHTML]="jobDescription.title | safeHtml"></h2>
             </div>
 
             <!-- Languages -->
-            <jr2-languages [languages]="job.jobContent.languageSkills"></jr2-languages>
+            <jr2-languages [languages]="job.jobContent.languageSkills"
```
birom: I think this should be also wrapped with 
```
<div class="detail__content">
```
element.

src/main/webapp/app/job-search/job-detail/job-detail.scss
```diff
@@ -25,6 +25,31 @@ td:first-of-type {
     margin-bottom: auto;
 }
 
+:host /deep/ .bullet {
+    .detail__content__title{
```
birom: I think we do not need this if the ``jr2-languages`` component is placed inside ``<div class="detail__content">`` an element. But we have to clarify this it Ingo, because on the screenshot attached to the issue has a slightly different content title as at other places (for example on the candidate detail page).

src/main/webapp/app/job-search/job-detail/job-detail.component.html
```diff
@@ -129,11 +129,13 @@ <h2 class="text--light" [innerHTML]="jobDescription.title | safeHtml"></h2>
...
+                           [containerClass]="'detail__content bullet'"
+                           [titleClass]="'detail__content__title'"></jr2-languages>
 
             <br/>
             <!-- Application -->
-            <div *ngIf="job.jobContent.applyChannel" class="detail__content">
+            <div *ngIf="job.jobContent.applyChannel" class="detail__content bullet">
```
birom: The [jobroom2 styleguide ](https://github.com/alv-ch/styleguide) projekt and the jobroom styles itsef have been developed using the [Block, Element, Modifier](http://getbem.com/naming/) methodololgy. According to this, I think the ``content__item--bullet`` class on the ``content __item`` elements would be a better solution.

e.g.:

```
 <div *ngIf="job.jobContent.applyChannel.formUrl" class="content__item content__item--bullet`">
                    <strong>{{ 'job-detail.application.online' | translate }}</strong>
                    <span>
                        <a href="{{ job.jobContent.applyChannel.formUrl }}">{{ 'job-publication-details.application.online-form' | translate}}</a>
                    </span>
                </div>
```

src/main/webapp/app/job-search/job-detail/job-detail.component.html
varusakov: It's already wrapped with `detail__content` class

src/main/webapp/app/job-search/job-detail/job-detail.component.html
varusakov: Thank you @birom, I wasn't aware of this notation.

src/main/webapp/app/job-search/job-detail/job-detail.component.html
birom: True, you set it with the ``containerClass`` attribute.

PR #725, opened 2018-08-16
JR2-1269 Escape markdown syntax on job-ad search page

varusakov:


package.json
```diff
@@ -39,6 +39,7 @@
     "ngx-webstorage": "2.0.1",
     "nouislider": "^10.1.0",
     "reflect-metadata": "0.1.10",
+    "remove-markdown": "^0.3.0",
```
pavax: i'd use exact versioning: "0.3.0"

PR #731, opened 2018-08-23
JR2-1073 Set permanent employment to false when short-term is selected

varusakov:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -360,6 +341,41 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
...
+                    }, 0);
+                }
+            });
+        this.updateEmploymentStartDateRelatedField();
+        this.jobPublicationForm.get('employment.shortEmployment').valueChanges
+            .takeUntil(this.unsubscribe$)
+            .subscribe((isShortEmployment: boolean) => {
+                this.jobPublicationForm.get('employment.employmentEndDate.permanent').setValue(!isShortEmployment);
+
+                const employmentEndDate = this.jobPublicationForm.get('employment.employmentEndDate.date');
```
pavax: why do we do that?

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -360,6 +341,41 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
...
+                    setTimeout(() => {
+                        this.employmentEndDateElementRef.nativeElement.focus();
+                    }, 0);
+                }
+            });
+        this.updateEmploymentStartDateRelatedField();
+        this.jobPublicationForm.get('employment.shortEmployment').valueChanges
+            .takeUntil(this.unsubscribe$)
+            .subscribe((isShortEmployment: boolean) => {
+                this.jobPublicationForm.get('employment.employmentEndDate.permanent').setValue(!isShortEmployment);
```
cjpsycos: empty also the employment.employmentEndDate.date

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
varusakov: screen shot 2018-08-23 at 15 23 46
In this case and date becomes required. If we choose short employment, date input will be hidden and we fail to submit the form

PR #732, opened 2018-08-23
Bugfix/jr2 1285

fabiangotzen:


src/main/webapp/app/shared/company/current-selected-company.service.ts
```diff
@@ -96,7 +96,7 @@ export class CurrentSelectedCompanyService {
                     companyId: company.id,
                     companyName: company.name,
                     companyStreet: company.street,
-                    companyHouseNr: '',
+                    companyZipCode: company.zipCode,
```
cjpsycos: Why removing companyHouseNr?
Why not use CompanyContactTemplate constructor?

PR #756, opened 2018-09-27
Feature/df 40

pavax:


src/main/resources/config/application.yml
```diff
@@ -161,6 +161,26 @@ jhipster:
...
+
+gateway:
+    rate-limiting:
+        enabled: true
+        rate-filter-options:
+            # Avoid fetching mass-fetching of sensitive candidate information
+            -
+                bucket-prefix: 'candidate-details'
+                method: 'GET'
+                url: '/candidateservice/api/candidates/*'
```
birom: I think the url should be ``/candidateservice/api/candidates/profiles/*``.
The candidate detail page fetches the candidate data with links like:
``https://www.job-room.ch/candidateservice/api/candidates/profiles/458d162d-fcb4-442b-829c-900e9f72518e?_ng=ZGU=``

PR #757, opened 2018-09-28
[DF-101] Show no. of equivalent positions in dashboard

wwebec:


src/main/webapp/app/dashboard/job-publication-detail/job-publication-detail.component.html
```diff
@@ -78,6 +78,9 @@ <h4 class="text--light" *ngIf="jobAdvertisement.status">
                 <small *ngIf="jobAdvertisement.publication.startDate" class="job-detail__date">
```
birom: It looks good, but would it be possible to add some extra margin?
"dashboard" detail page:
selection_098

public detail page:
selection_099

I think we can use the css classes from the public detail page.
"dashboard" detail: ``job-detail__title`` > ``job-detail__date``
"public" detail page: ``detail__title`` >  ``detail__date``
 
thx

wwebec: done

PR #765, opened 2018-10-02
DF-154 Preserve html tags during markdown cleanup

varusakov:
This code do same things as `remove-markdown` library except for removing HTML tags from text.


src/main/webapp/app/shared/pipes/markdown-escape.pipe.ts
```diff
@@ -1,17 +1,59 @@
...
 
     transform(value: any): any {
         if (value) {
-            return removeMd(value)
+            return this.escapeMarkdown(value)
         }
         return value;
     }
 
+    escapeMarkdown(value: any) {
```
birom: Please add unit tests.

varusakov: Added unit tests

PR #768, opened 2018-10-04
DF-159 Return null in case of error

varusakov:


src/main/webapp/app/candidate-search/services/candidate.service.ts
```diff
@@ -90,7 +90,7 @@ export class CandidateService {
                     return this.http.get<Candidate>(`${this.resourceUrl}/${candidateProfile.id}`);
                 }
                 return Observable.of(null as Candidate);
-            });
+            }).catch(() => Observable.of(null as Candidate));
```
pavax: this method is called from the CandidateDetailComponent ngOnit component, so shouldn't that one handle the error case?

varusakov: Done. I also found that we have similar issue with candidate's job center

PR #781, opened 2018-10-11
added owasp dep check incl. travis build integration

ivanmrva:


gradle/owasp-dep-check.gradle
```diff
@@ -0,0 +1,35 @@
...
+        classpath 'org.owasp:dependency-check-gradle:3.3.2'
+    }
+}
+
+apply plugin: 'org.owasp.dependencycheck'
+
+apply plugin: 'maven-publish'
+
+publishing {
+    repositories {
```
birom: Why do we need this if we already have something similar in the ``maven-publish.gradle``?

birom: We already have this in the ``maven-publish.gradle``. Is not it possible to reuse the ``repositories`` section?

PR #784, opened 2018-10-19
DF-282: Build and push docker image by travis

birom:
-add pushDockerImage task to the travis build pipeline


gradle/docker.gradle
```diff
@@ -8,23 +10,38 @@ buildscript {
...
+        url = 'alvch-dockerv2-local.jfrog.io'
+        username = System.env.ARTIFACTORY_USERNAME
+        password = System.env.ARTIFACTORY_PASSWORD
+    }
+}
 
 import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+import com.bmuschko.gradle.docker.tasks.image.DockerPushImage
 
 task copyDockerFiles(type: Copy) {
```
ivanmrva: I think if you just would use the relative path to the built artifact (WAR) in the Dockerfile, you wouldn't need this copy task at all. Anyway since the Dockerfile is in the project I think it is OK to reference from it files at the exact location as they are. That way you can even build the docker image locally on the CLI and don't have to first copy the WAR somewhere else.

birom: I agree that your idea is cleaner. 
Unfortunaetly the suggestion does not work now, due to a bug in the [docker-java](https://github.com/docker-java/docker-java/) project.

Problem:
We need a .dockerignore file do exclude the unwanted files from the docker build context. Otherwise the build takes too long.
With the native docker client the following .dockerignore file works fine:
```
# https://docs.docker.com/engine/reference/builder/#dockerignore-file
# by default ignore everything except the jar file
**/*
!build/libs/*
```

But the gradle plugin fails with the following error message:
```
Starting a Gradle Daemon, 1 incompatible and 1 stopped Daemons could not be reused, use --status for details

> Configure project : 
build/libs/jobroom-1.0.0-SNAPSHOT.war
/home/mabi/development/projects/seco/jobroom2

> Task :buildDockerImage 
Building image using context '/home/mabi/development/projects/seco/jobroom2'.
Using tags 'alvch-dockerv2-local.jfrog.io/jobroom:latest', 'alvch-dockerv2-local.jfrog.io/jobroom:1.0.0-SNAPSHOT' for image.

Step 1/6 : FROM openjdk:8-jre-alpine

---> b1bd879ca9b3
Step 2/6 : ENV SPRING_OUTPUT_ANSI_ENABLED=ALWAYS     SLEEP=0     JAVA_OPTS=""

---> Using cache
---> a4a2c51ac229
Step 3/6 : ARG JAR_FILE

---> Using cache
---> e790db919ceb
Step 4/6 : COPY ${JAR_FILE} /jobroom.war


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':buildDockerImage'.
> Could not build image: COPY failed: stat /var/lib/docker/tmp/docker-builder554848750/build/libs/jobroom-1.0.0-SNAPSHOT.war: no such file or directory

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 10s
1 actionable task: 1 executed
```
I found a bug report in the docker-java project that is used by the gradle plugin:
https://github.com/docker-java/docker-java/issues/999

Until this is not solved, I would use the current approach.

PR #791, opened 2018-10-29
DF-304 Implement capitalization pipe

varusakov:


src/main/webapp/app/candidate-search/candidate-detail/candidate-detail.component.html
```diff
@@ -74,7 +74,7 @@ <h4 class="detail__content__title candidate-detail__content__title--occupation">
                             {{ (jobExperience.lastJob ? 'candidate-detail.job-experience.last-job.label' :
                             'candidate-detail.job-experience.job.label') | translate}}
                         </strong>
-                        <span class="text--bold">{{ jobExperience.occupationLabel }}</span>
+                        <span class="text--bold">{{ jobExperience.occupationLabel | jr2Capitalize }}</span>
```
birom: We already have the [JhiCapitalizePipe](https://github.com/jhipster/ng-jhipster/blob/master/src/pipe/capitalize.pipe.ts) in the project. I would prefer to use that one.

varusakov: @birom JhiCapitalizePipe makes all string lowercase and than capitalize first letter. We have occupations that contains capital chars in the middle. That's why I've implemented own pipe

birom: Ok, I see. Thank you!

PR #796, opened 2018-11-02
Feature/df 242 job publication post address fields

99999989:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-mapper.ts
```diff
@@ -249,6 +261,65 @@ export class JobPublicationMapper {
...
+
+    static mapPostAddressToCreateJobAdvertisement(postAddress, jobAd: CreateJobAdvertisement):  void {
+
+        jobAd.applyChannel.postAddress.name = postAddress.paperAppCompanyName;
+        jobAd.applyChannel.postAddress.countryIsoCode = postAddress.paperAppCountryCode;
+
+        const postalCode = postAddress.paperAppZip.zip;
+        const city = postAddress.paperAppZip.city;
+
+        if (postAddress.paperAppPostboxNr) {
```
cjpsycos: What's happen if i have only name, postalCode and ctiy?

src/main/webapp/app/home/tools/job-publication-tool/job-publication-mapper.ts
```diff
@@ -249,6 +261,65 @@ export class JobPublicationMapper {
...
+            jobAd.applyChannel.postAddress.postOfficeBoxNumber = postAddress.paperAppPostboxNr;
+            jobAd.applyChannel.postAddress.postOfficeBoxPostalCode = postalCode;
+            jobAd.applyChannel.postAddress.postOfficeBoxCity = city;
+        } else if (postAddress.paperAppStreet) {
```
cjpsycos: It's just a "else" block, so I can also provide only name, postalCode and ctiy

leanoo:
@cjpsycos the cases where job-ads are copied from dashboard, etc. are managed and working properly, since they all use the job-publication-mapper.ts

PR #812, opened 2018-11-23
[DF-379] Fix validation of the job publication page: added error messages of max length validation

wwebec:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.html
```diff
@@ -502,8 +504,9 @@
                             </select>
                             <jr2-input-error-message [formGroup]="jobPublicationForm"
                                                      path="contact.salutation"
-                                                     [errorCodes]="['required']"
-                                                     [keys]="['home.tools.job-publication.messages.validate.required']">
+                                                     [errorCodes]="['required','maxlength']"
```
birom: It is a select box with predefined values. How can be the value longer than excepted?

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.html
```diff
@@ -606,8 +613,9 @@
                             </select>
                             <jr2-input-error-message [formGroup]="jobPublicationForm"
                                                      path="publicContact.salutation"
-                                                     [errorCodes]="['required']"
-                                                     [keys]="['home.tools.job-publication.messages.validate.required']">
+                                                     [errorCodes]="['required','maxlength']"
```
birom: The same as my other comment.

PR #818, opened 2018-12-05
Feature/df 350 agb validation

birom:


src/main/resources/config/liquibase/changelog/20181120080000_add_legal_terms_to_user_info.xml
```diff
@@ -0,0 +1,12 @@
...
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+    <changeSet id="20181120080000-1" author="mabi">
+        <addColumn tableName="user_info">
+            <column name="legal_terms_accepted_at" type="datetime"/>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
```
cjpsycos: also add an migration script for filling this field. (should be filled if this field is null and status is registered)

PR #821, opened 2018-12-05
DF-400 Made company zip code optional when street is not empty

varusakov:


src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -418,6 +420,30 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
...
+        const streetControl = companyControl.get('street');
+
+        streetControl.valueChanges
+            .takeUntil(this.unsubscribe$)
+            .subscribe((street) => {
+                this.companyZipCodeOptional = !!street
+            });
+
+        companyControl.valueChanges
+            .filter(() => !!companyControl.get('zipCode'))
```
fabiangotzen: Hmm, shouldn't this be postboxNumber? When I read the description I would expect the same behaviour as we have in application channels -> by Post.  When I run it locally there is no change to the street validator when i enter a PO box no. in the company section.
Please clarify this with @leanoo , maybe i misunderstood.

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
```diff
@@ -418,6 +420,30 @@ export class JobPublicationToolComponent implements OnInit, OnDestroy {
...
+            .map(() => companyControl.get('zipCode').valid)
+            .distinctUntilChanged()
+            .takeUntil(this.unsubscribe$)
+            .subscribe((zipCodeValid) => {
```
fabiangotzen: Also, if this should be the case please also check the job-publication-mapper, f.E
```

        if (jobPublicationForm.company.postboxZipCode) {
            Object.assign(jobAd.company, {
                postOfficeBoxPostalCode: jobPublicationForm.company.postboxZipCode.zip,
                postOfficeBoxCity: jobPublicationForm.company.postboxZipCode.city
            });
        }
```

might be that we will send some incomplete data to the backend

src/main/webapp/app/home/tools/job-publication-tool/job-publication-tool.component.ts
varusakov: PO box no. is already optional field. I don't think that we need to change validation for it somehow. @tonim42 said that this issue is related to zip code field.
screen shot 2018-12-04 at 14 11 11

PR #829, opened 2018-12-12
DF-288 Clean up current user after session is expired

varusakov:


src/main/webapp/app/shared/auth/auth-jwt.service.ts
```diff
@@ -3,13 +3,23 @@ import { HttpClient } from '@angular/common/http';
...
+
+    static readonly TOKEN = 'authenticationToken';
+
+    private timeout;
+
     constructor(
         private http: HttpClient,
         private $localStorage: LocalStorageService,
-        private $sessionStorage: SessionStorageService
+        private $sessionStorage: SessionStorageService,
```
fabiangotzen: like discussed, just a small cosmetic thing, but we never use a $-Prefix elsehwere in jobroom, so if there is no specific reason for it i'm missing we can remove it.

PR #832, opened 2018-12-12
DF-314 Add title before personal message in email

varusakov:


src/main/resources/i18n/messages_fr.properties
```diff
@@ -19,8 +19,10 @@ email.creation.text1=Votre compte Job-Room a été créé avec le nom d''utilisa
 # Reset email
 email.reset.title=Réinitialisation de mot de passe Job-Room
 email.reset.text1=Un nouveau mot de passe pour votre compte Job-Room ''{0}'' a été demandé. Veuillez cliquer sur le lien ci-dessous pour le réinitialiser :
-email.anonymousContact.mail-subject=Demande de contact d'un employeur à travers Job-Room
 # Anonymous contact email
+email.anonymousContact.mail-subject=Demande de contact d'un employeur à travers Job-Room
+email.anonymousContact.body=L'employeur ci-après a consulté votre profil dans Job-Room et souhaiterait que vous le recontactiez au moyen des coordonnées ci-dessous pour discuter les prochains pas.
```
fabiangotzen: Just a heads up: I think the single quatation might cut of the string here.  We should run a quick test after its deployed on staging

PR #833, opened 2018-12-12
Feature/df 349 check and accept agb ui

birom:
selection_131
Mobile:
selection_133


src/main/webapp/app/account/registration/questionnaire/registration-questionnaire.component.ts
```diff
@@ -10,10 +12,14 @@ import { RegistrationDialogService } from '../registration-dialog.service';
...
     public roleForm: FormGroup;
+    public legalTermsUrl$: Observable<string>;
 
     constructor(private fb: FormBuilder,
                 private backgroundUtils: BackgroundUtils,
-                private registrationDialogService: RegistrationDialogService) {
+                private registrationDialogService: RegistrationDialogService,
+                private legalTermsService: LegalTermsService) {
+        this.legalTermsUrl$ = this.legalTermsService.getCurrentLegalTermsUrl()
+            .catch((error) => Observable.of('error'));
```
varusakov: I think we can move catch clause to service instead of repeating it in components

birom: I've done like that intentionally. I think the component is responsibe for the error handling in this case.

varusakov: I see your point, txn

src/main/webapp/app/account/registration/legal-terms/legal-terms.component.html
```diff
@@ -0,0 +1,36 @@
+<ng-template>
+    <ng-container *ngIf="legalTermsUrl$ | async as legalTermsUrl">
+        <div class="modal-header">
+            <h4 class="modal-title">{{ 'global.legal-terms-dialog.header.title' | translate }}</h4>
+
+        </div>
+        <div class="modal-body">
+            <ul class="alert alert-info">
+                <li>{{ 'global.legal-terms-dialog.header.content1' | translate }}</li>
```
cjpsycos: why refers the key to the dialog header? it's in the body.
why 2 contents?

src/main/webapp/app/account/account.module.ts
```diff
@@ -20,6 +20,8 @@ import { RegistrationDialogService } from './registration/registration-dialog.se
 import { RegistrationAccessCodeComponent } from './registration/access-code/registration-access-code.component';
 import { RegistrationGuardService } from './registration/registration-guard.service';
 import { ContactTemplateManagementComponent } from '../account/contact-template-management/contact-template-management.component';
+import { LegalTermsComponent } from './registration/legal-terms/legal-terms.component';
```
cjpsycos: why are these in registration? legal-terms is on the account.

src/main/webapp/app/account/registration/legal-terms/legal-terms.component.ts
```diff
@@ -0,0 +1,79 @@
...
+        this.loginService.logout();
+        if (this.noEiam) {
+            this.router.navigate(['']);
+        } else {
+            document.location.href = 'authentication/logout';
+        }
+    }
+
+    acceptLegalTerms(): void {
+        this.registrationService.acceptLegalTerms()
```
cjpsycos: shouldn't it be a legal-terms-service?

src/main/webapp/app/account/registration/legal-terms/legal-terms.component.html
birom: Thx, the header is wrong. On Ingos mockup there is a list of infos. I did not want to put html into the translations property files. I can change it it, it would be more flexible.

PR #836, opened 2018-12-13
DF 422 logging enhanced for end mail

fankhauserf:


src/main/java/ch/admin/seco/jobroom/service/MailService.java
```diff
@@ -167,6 +167,7 @@ private void sendEmail(String to, String subject, String content, boolean isMult
             javaMailSender.send(mimeMessage);
             log.info("Successfully sent email to '{}' with subject '{}'", to, subject);
         } catch (MessagingException e) {
+            log.info("Sending email failed for '{}'", to, e);
```
cjpsycos: log level should be "error"

PR #839, opened 2018-12-14
DF-347: Fix logout

birom:


src/main/java/ch/admin/seco/jobroom/config/SecurityConfiguration.java
```diff
@@ -269,13 +268,11 @@ private SamlAuthenticationFailureHandler authenticationFailureHandler() {
...
-                    UserPrincipal principal = (UserPrincipal) authentication.getPrincipal();
-                    applicationEventPublisher.publishEvent(BusinessLogEvent.of(USER_LOGOUT)
-                        .withObjectType(USER.typeName())
-                        .withObjectId(principal.getId().getValue())
-                        .withAuthorities(principal.getAuthoritiesAsString())
-                    );
+                public void onLogoutSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication)
+                    throws IOException, ServletException {
+
+                    applicationEventPublisher.publishEvent(BusinessLogEvent.of(USER_LOGOUT).withObjectType(USER.typeName()));
```
pavax: mhhh interesting.... wondering why it didn't work....

birom: Hi @pavax, 

at this point the  authentication is already ``null``. 
If we really want this feature, I do not see any other option than add a new custom ``logoutHandler`` to the SAMLLogoutFilter here: https://github.com/alv-ch/jobroom2/blob/develop/src/main/java/ch/admin/seco/jobroom/security/saml/infrastructure/dsl/SAMLConfigurer.java#L374.
Do you have an other idea?

Thank you

pavax: I'd not hook myself into that part since the session has not been destroyed at there.... at least not yet... I'd leave it as it is right now....

birom: Ok, thx. I am going to merge it.

PR #847, opened 2019-01-03
Bugfix: Let's try to use a guard if we activate the landing page so t…

pavax:
…hat at the end we don't get to see that dummy component's content as a user


src/main/webapp/app/shared/landing-page/landing-page.component.ts
```diff
@@ -9,13 +7,10 @@ import { Principal } from '..';
 })
 export class LandingPageComponent implements OnInit {
 
-    constructor(private landingPageService: LandingPageService,
-                private principal: Principal) {
+    constructor() {
```
wwebec: there is no need for empty constructor

src/main/webapp/app/shared/landing-page/landing-page.component.ts
```diff
@@ -9,13 +7,10 @@ import { Principal } from '..';
...
     }
 
     ngOnInit() {
```
wwebec: there is no need for empty lifecycle hook

src/main/webapp/app/shared/landing-page/landing-page.component.ts
```diff
@@ -9,13 +7,10 @@ import { Principal } from '..';
 })
 export class LandingPageComponent implements OnInit {
```
wwebec: if empty lifecycle hook were removed,  the OnInit could be removed as well

src/main/webapp/app/shared/landing-page/landing-page.guard.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Principal } from '..';
+import { LandingPageService } from './landing-page.service';
+
+@Injectable()
+export class LandingPageGuard implements CanActivate {
+
+    constructor(private principal: Principal, private landingPageService: LandingPageService) {
+    }
+
+    canActivate(next: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
```
wwebec: what about simplifying the above to 
```
    canActivate(next: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean {
        this.principal.identity()
            .then((currentUser) => this.landingPageService.navigateToLandingPage(currentUser))
            .catch(() => this.landingPageService.navigateToLandingPage()
        );
        return false;
    }
```

PR #862, opened 2019-02-25
DF-729: SYSADMIN should also be able to contact a candidate

pavax:


pavax:
SYSADMIN should not be able to contact a candidate since he doesn't have a associated company

PR #872, opened 2019-04-25
DF-921: cleanup properties

leanoo:


gradle.properties
```diff
@@ -16,7 +16,7 @@ liquibase_hibernate5_version=3.6
 jacoco_version=0.7.9
 springfox_swagger2_version=2.9.2
 spring_security_saml2_version=1.0.9.RELEASE
-alv_shared_library_version=0.0.1-build-74
+alv_shared_library_version=0.0.1-build-77
```
birom: We need to upgrade the the latest version after this is merged: https://github.com/alv-ch/alv-shared-library/pull/12

PR #875, opened 2019-05-23
Feature/df 970

dvoracek-martin:


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoRole.java
```diff
@@ -0,0 +1,7 @@
+package ch.admin.seco.jobroom.domain;
+
+public enum UserInfoRole {
+    JOB_SEEKER,
+    COMPANY,
+    PAV
+}
```
fabiangotzen: I've noticed that this would be third enum declaration of Roles in the gateway, apart from EiamClientRole.java & UserInfoResource.Role. 

As a suggestion, I think we could remove this UserInfoRole and move the Role Set from the UserInfoResource to domain.enumeration as Role.java (Since the Role isn't really linked to the UserInfo) and use it. What do you think?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/impl/messaging/UserInfoEventGateway.java
```diff
@@ -0,0 +1,53 @@
...
+public class UserInfoEventGateway {
+
+    private final MessageChannel userInfoEventChannel;
+    private final String EVENT = "event";
+    private final String RELEVANT_ID = "relevant-id";
+    private final String PAYLOAD_TYPE = "payload-type";
+    private final String PARTITION_KEY = "partitionKey";
+
+    public UserInfoEventGateway(MessageBrokerChannels messageBrokerChannels) {
+        userInfoEventChannel = messageBrokerChannels.userInfoEventChannel();
```
birom: Do not use directly the Output bindings. At other places eg.: in the job-ad-serivce we put the messages at first into a persistent ``JdbcChannelMessageStore``. Please have a look at this branch: https://github.com/alv-ch/jobroom2/tree/feature/DF-970-mabi

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserRegisteredEvent.java
```diff
@@ -0,0 +1,25 @@
+package ch.admin.seco.jobroom.domain;
+
+public class UserRegisteredEvent extends AbstractUserEvent {
+
+    private final UserInfoRole userInfoRole;
+    private final Long personNumber;
```
cjpsycos: PersonNumber should be available also in UserUnregisteredEvent.
Move this in the AbstractUserEvent

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoRole.java
cjpsycos: Call it simple Role

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/impl/messaging/MessageHeaders.java
```diff
@@ -0,0 +1,15 @@
...
+public final class MessageHeaders {
+
+    private MessageHeaders() {
+
+    }
+
+    public static final String EVENT = "event";
+    public static final String SOURCE_SYSTEM = "sourceSystem";
+    public static final String PARTITION_KEY = "partitionKey";
+    public static final String PERSON_NUMBER = "personNumber";
```
cjpsycos: This should be just a payload

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/impl/messaging/MessageBrokerChannels.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.annotation.Output;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.SubscribableChannel;
+
+public interface MessageBrokerChannels {
+
+    String CANDIDATE_EVENTS = "candidateEvents";
+
+    String USER_INFO_EVENTS = "userInfoEvents";
```
cjpsycos: This would be USER_EVENTS when the topic is called "user.event"

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/impl/messaging/MessageHeaders.java
pavax: there should not be a PERSON_NUMBER in the Headers

PR #878, opened 2019-06-05
DF-979 - remove candidate email

fabiangotzen:


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/dto/CandidateAnonymousContactInfo.java
```diff
@@ -0,0 +1,26 @@
+package ch.admin.seco.jobroom.service.dto;
+
+public class CandidateAnonymousContactInfo {
```
birom: Rename it to ``CandidateContactInfo`` as in the candidate service.

fabiangotzen: done

PR #884, opened 2019-07-10
Df 1070/remove anonymous contact

fabiangotzen:
relates to. https://github.com/alv-ch/candidate-service/pull/127


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/web/rest/EmailResource.java
```diff
@@ -1,35 +0,0 @@
-package ch.admin.seco.jobroom.web.rest;
```
birom: Remove the corresponding configuration form the ``AbstractSecurityConfig``, please.

fabiangotzen: Thanks, missed that. I'll move it to candidate as well

PR #888, opened 2019-08-28
Feature/df 953 poc

pavax:


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/config/togglz/FeatureCodeListActivationStrategy.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public String getId() {
+        return "feature-code-list";
+    }
+
+    @Override
+    public String getName() {
+        return "Feature Code Liste";
```
dvoracek-martin: German is used here intentionally or it's just a typo in the word "Liste"?

pavax: yes it's intentionally ... if we'd have a ui (togglz-console) to manage the toggle's that label would appear

PR #889, opened 2019-09-03
Insert or update company

99999989:
I've replaced the existing logic with a custom "upsert" logic using `findByExternalId()`


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
```diff
@@ -215,39 +215,29 @@ private boolean isPersonNumberAlreadyTaken(Long personNumber) {
...
         Optional<Company> companyInDb = this.companyRepository.findByExternalId(getFullUid(firm.getUidPrefix(), firm.getUid()));
-        if (companyInDb.isPresent()) {
-            return companyInDb.get();
-        } else {
-            Company company = toCompany(firm);
-            return this.companyRepository.save(company);
-        }
+        Company company = toCompany(firm);
+        companyInDb.ifPresent(c -> company.setId(c.getId()));
+        return this.companyRepository.save(company);
```
fabiangotzen: I think we could simplify this to: (same for the other method)
```
    private Company storeCompany(FirmData firm) {
        Company company = toCompany(firm);
        this.companyRepository.findByExternalId(getFullUid(firm.getUidPrefix(), firm.getUid()))
            .ifPresent(companyInDb -> company.setId(companyInDb.getId()));
        return this.companyRepository.save(company);
    }

@pavax Personally I don't really like that we have different approaches to our domain entities inside this service: Some Entities are pure Datacontainers like Company with public setters while others are designed with a more domain driven approach in mind like BlackListedAgent or UserInfo. But probably out of scope, so I would suggest to create a technical debt task.

File: jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java [hunk 2]
```diff
@@ -215,39 +215,29 @@ private boolean isPersonNumberAlreadyTaken(Long personNumber) {
...
         Optional<Company> companyInDb = this.companyRepository.findByExternalId(organization.getExternalId());
-        if (companyInDb.isPresent()) {
-            return companyInDb.get();
-        } else {
-            Company company = toCompany(organization);
-            return this.companyRepository.save(company);
-        }
+        Company company = toCompany(organization);
+        companyInDb.ifPresent(c -> company.setId(c.getId()));
+        return this.companyRepository.save(company);
```
fabiangotzen: Please also have a look at the failing Tests in the RegistrationServiceTest: insertNewAgentWithExistingCompany and insertNewEmployerwithExistingCompany. 

Boyscout Proposal: I don't think we use the Builder-style Setters inside the Company Entity anywhere, so I think we could remove them, f.E
```
    public Company source(CompanySource source) {
        this.source = source;
        return this;
    }

Comment (99999989) [reply jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java]: I've fixed the tests and removed the builder style setters.

Comment (fabiangotzen) [reply jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java]: Previously we had no interaction with the database when we had an existing company, because we interact with the database now (since we update it) we need to mock this out and return a dummyCompany.
Otherwise we'll throw an exception in Userinfo.java since we need a Company to create an Accountability

```
    void addCompany(Company company) {
        Accountability accountability = new Accountability(AccountabilityType.USER, company);
```

So by doing this 
```
when(mockCompanyRepository.save(any())).thenReturn(getDummyCompany());
```
and this: 
```
        verify(mockCompanyRepository, atLeastOnce()).save(any());
```
the test should run through

99999989: Thanks for the hint. It's fixed now.

PR #901, opened 2020-02-10
Update jenkinsfile.groovy

cinhtau:
Display logs and get result object, if build has no errors return true, else false will abort deployment
Although it is scripted, basically not needed, just for clarity in this PR

See also https://github.com/openshift/jenkins-client-plugin#actions-speak-louder-than-words


openshift/jobroom2/jenkinsfile.groovy
```diff
@@ -129,11 +129,14 @@ pipeline {
                                 "-p", "NAMESPACE=jobroom-dev",
                             ]))
 
-                            openshift.selector('bc', 'jobroom-gateway-app-docker').startBuild("--from-dir .").watch {
-                                if(it.object().status.phase == 'Complete') {
-                                    return true
-                                }
+                            def build = openshift.selector('bc', 'jobroom-gateway-app-docker').startBuild("--from-dir .")
```
birom: Deployment in the next step started a bit early without the watch. This resulted that, not the latest container image was deployed. Does this change also fix that problem?

cinhtau: Yes. It is in the later code.

birom: Ok, thx.

PR #909, opened 2020-03-27
Feature/df 1979 from develop automatic deativation nicht eingeloggter nutzer

hamidsoleymani:


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoRepository.java
```diff
@@ -29,4 +33,7 @@
 
     @Query("select u from UserInfo u where u.registrationStatus = 'UNREGISTERED' order by u.id")
     Page<UserInfo> findAllUnregisteredUsers(Pageable pageable);
+
+    @Query("select u from UserInfo u where u.lastLoginAt < :todayMinusInactiveDays  and u.stesInformation.personNumber is null and u.registrationStatus in :registrationStatus")
```
emina-str: I think `:registrationStatus` won't work, you are giving enum values. You should probably just add string values in the query, and discard list of registration statuses from parameters.

the name of the method should be something like `findAllPavAndAgByLastLoginAtBefore`

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
```diff
@@ -320,7 +307,14 @@ private void doUnregister(String searchParam, SearchParamType searchParamType, E
         LOGGER.info("JobSeeker with {} : {} has been unregistered", searchParamType, searchParam);
         applicationEventPublisher.publishEvent(BusinessLogEvent.of(USER_UNREGISTERED).withObjectId(userInfo.getId().getValue()).withAdditionalData("affectedRole", role.getName()));
         applicationEventPublisher.publishEvent(new UserUnregisteredEvent(userInfo, toUserInfoRole(role)));
+    }
 
+    public void doUnregister(UserInfo userInfo, EiamClientRole role) throws UserNotFoundException {
+        userInfo.unregister();
```
emina-str: this part is the same as above method.
```
userInfo.unregister();
this.eiamAdminService.removeRole(userInfo.getUserExternalId(), role);
LOGGER.info(role.getName() + " with email: {} has been unregistered", userInfo.getEmail());        applicationEventPublisher.publishEvent(BusinessLogEvent.of(USER_UNREGISTERED).withObjectId(userInfo.getId().getValue()).withAdditionalData("affectedRole", role.getName()));
applicationEventPublisher.publishEvent(new UserUnregisteredEvent(userInfo, toUserInfoRole(role)));
```
 So for above, just use code : 
`doUnregister(userInfo, role)`

and in this method will be implementation.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoRepository.java
hamidsoleymani: 1) I changed the method name to "findAllPavAndAgByLastLoginAtBefore".
2) I tested :registrationStatus with enum values and it works fine. But I changed the Query as you proposed, since it will be easier to read.

jobroom-gateway-app/src/main/resources/config/application.yml
```diff
@@ -221,3 +221,10 @@ togglz:
         BU:
             enabled: true
             strategy: feature-code-list
+
+jobroom:
```
emina-str: you added also this parameters to central-config, but if all are the same for each environment (for now), it is not needed there. then we just take it from here. lets say it's optional, you can remove it from central-config, but it doesn't hurt if it stay ...

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+        this.userInfoRepository = userInfoRepository;
+        this.mailService = mailService;
+        this.unregistrationConfig = unregistrationConfig;
+    }
+
+    @Transactional
+    @Scheduled(cron = "${jobroom.user.unregistration.cron}")
+    public void reportCurrentTime() {
+        LOGGER.info("Started PAV and AG unregistration:");
+        if (unregistrationConfig.isShouldRun()) {
```
emina-str: In my opinion, this whole logic belongs to `RegistrationService`, and not in this task class.
I would here just call a method from RegistrationService, where all implementation would be.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+            }
+        }
+    }
+
+    private void sendEmail(UserInfo userInfo, String role) {
+        EmailDTO emailDTO = createEmailDTO(userInfo, role);
+        mailService.sendUserUnregistrationEmail(emailDTO);
+    }
+
+    private EmailDTO createEmailDTO(UserInfo userInfo, String role) {
```
emina-str: there is a `\src\main\java\ch\admin\seco\jobroom\service\mapper` package, where this kind of mappings belong.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+    public void reportCurrentTime() {
+        LOGGER.info("Started PAV and AG unregistration:");
+        if (unregistrationConfig.isShouldRun()) {
+            Pageable nextPageable = PageRequest.of(0, 50);
+            Page<UserInfo> pageResult;
+            do {
+                LocalDateTime todayMinusInactiveDays = LocalDateTime.now().minusDays(unregistrationConfig.getUnregisterUserAfterInactivityDays());
+                pageResult = userInfoRepository.findAllPavAndAgByLastLoginAtBefore(todayMinusInactiveDays, nextPageable);
+                pageResult.forEach(userInfo -> {
+                    try {
```
emina-str: `try-catch` is very time consuming. I think it's better to avoid it.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Transactional
+    @Scheduled(cron = "${jobroom.user.unregistration.cron}")
+    public void reportCurrentTime() {
+        LOGGER.info("Started PAV and AG unregistration:");
+        if (unregistrationConfig.isShouldRun()) {
+            Pageable nextPageable = PageRequest.of(0, 50);
+            Page<UserInfo> pageResult;
+            do {
+                LocalDateTime todayMinusInactiveDays = LocalDateTime.now().minusDays(unregistrationConfig.getUnregisterUserAfterInactivityDays());
```
emina-str: this should go above the `do-loop`. it's a constant for the task.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+                nextPageable = pageResult.nextPageable();
+            } while (pageResult.hasNext());
+            LOGGER.info("Finished PAV and AG unregistration:");
+        }
+    }
+
+    private void processUserInfo(UserInfo userInfo) throws UserNotFoundException {
+        if (hasText(userInfo.getEmail())) {
+
+            if (userInfo.getRegistrationStatus() == RegistrationStatus.VALIDATION_PAV) {
```
emina-str: What about registration status `REGISTERED` ?
There is a method in `EiamAdminService` called `getRoles(String userExternalId)` and I think you should utilize it.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+
+import javax.transaction.Transactional;
+import java.time.LocalDateTime;
+
+import static ch.admin.seco.jobroom.security.registration.eiam.EiamClientRole.ROLE_COMPANY;
+import static ch.admin.seco.jobroom.security.registration.eiam.EiamClientRole.ROLE_PRIVATE_EMPLOYMENT_AGENT;
+import static org.springframework.util.StringUtils.hasText;
+
+@Component
+public class UnregisterUsersScheduleTask {
```
emina-str: I think `@EnableScheduling` is missing? And then it's used with `@Configuration` annotation.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+            LOGGER.info("Finished PAV and AG unregistration:");
+        }
+    }
+
+    private void processUserInfo(UserInfo userInfo) throws UserNotFoundException {
+        if (hasText(userInfo.getEmail())) {
+
+            if (userInfo.getRegistrationStatus() == RegistrationStatus.VALIDATION_PAV) {
+                registrationService.doUnregister(userInfo, ROLE_PRIVATE_EMPLOYMENT_AGENT);
+                // sendEmail(userInfo, ROLE_PRIVATE_EMPLOYMENT_AGENT.toString());
```
emina-str: why still commented?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: @EnableScheduling is on class JobroomApp. That's why we don't need it

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: We only unregister PV and AG. We get from DB users who are registered and are PAV und AG. correct?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: correct!

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
hamidsoleymani: ok!

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: This is an option which has its own advantages, as you said. Otherwise this class encapsulates the unregistration tasks.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/JobroomApp.java
```diff
@@ -29,7 +30,7 @@
 import java.util.Collection;
 
 @SpringBootApplication
-@EnableConfigurationProperties({LiquibaseProperties.class, ApplicationProperties.class})
+@EnableConfigurationProperties({LiquibaseProperties.class, ApplicationProperties.class, UnregistrationConfig.class})
```
birom: Please rename ``UnregistrationConfig`` to ``UnregistrationProperties``.
The name of the ``@ConfigurationProperties`` annotated files have ``Properties`` suffix.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: I think in this case it is not necessary to move it to mapper package. What do you think?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: It will be done in coming sprint, since it was added on top of the current sprint. And we don't have any email text form Fach

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: I don't understand the call for getRoles.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: If no exception is thrown, try-catch has no time penalty and it is as fast as without try-catch.
I think we should not avoid language features because we think that we might save a few milliseconds of something that isn't slow in the first place.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/JobroomApp.java
hamidsoleymani: ok!

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
emina-str: ok.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
emina-str: But why you added in Repository for fetching all PAV and AG also the status `REGISTERED`?
Wouldn't there then be needed only this two statuses `VALIDATION_PAV`, `VALIDATION_EMP`?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
emina-str: ok.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
emina-str: ok.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: Because we are only interested in those users who are still registered. if we omitted status REGISTERED we would also fetch users who are UNREGISTERED and REGISTERED. But we are not interested in UNREGISTERED users

jobroom-gateway-app/src/main/resources/config/application.yml
hamidsoleymani: I will let is as is. ok?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
hamidsoleymani: > But why you added in Repository for fetching all PAV and AG also the status `REGISTERED`?
> Wouldn't there then be needed only this two statuses `VALIDATION_PAV`, `VALIDATION_EMP`?

I removed the status REGISTERED

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/unregistration/UnregisterUsersScheduleTask.java
```diff
@@ -0,0 +1,96 @@
...
+                return;
+            }
+            if (userInfo.getRegistrationStatus() == RegistrationStatus.VALIDATION_EMP) {
+                registrationService.doUnregister(userInfo, ROLE_COMPANY);
+                // sendEmail(userInfo, ROLE_COMPANY.toString());
+            }
+        }
```
tonim42: This logic is not correct. The registration status is not relevant.

hamidsoleymani:
> Hallo Hamid
> 
> jetzt verstehe ich Deine Fragen. Man weiss nicht, welche Rolle der user hat. In der User_Info-Tabelle steht das nicht. Man muss daher zuerst das eIAM nach der Rolle fragen und dann diese Rolle deregistrieren. Die Nutzer mit Status VALIDATION_EMP und VALIDATION_PAV haben noch keine Rolle. Ich muss noch mal anschauen, ob die Deregistrierung dieser Nutzer überhaupt möglich ist.

tonim42:
@hamidsoleymani @emina-str 
I have found there is a problem in the existing code. Class UserInfoResource lines 104 t o110, here the case NO_ROLE is treated like JOB_SEEKER, which results in a wrong BusinessLog entry.
This is a bit of a mess and I appreciate your help in getting this fixed.

In the case NO_ROLE, the EiamAdminService.removeRole() method is unnecessary, all that needs to be called is the UserInfo.unregister(), which removes the accountabilities and resets the registration status.

tonim42:
@seco-mgabor Für Dich auch relevant. Betrifft die Mechanismen des eIAM.

PR #913, opened 2020-04-08
DF-2090 add healthcheck

cinhtau:
- reformat yaml, since it had mixed indentation widths
- remove route, since it is not needed


openshift/jobroom2/jobroom-gateway-app/deployment-config.yml
```diff
@@ -2,96 +2,81 @@ kind: List
...
+                                    mountPath: "/eiam"
+                                    readonly: true
+                            resources:
+                                requests:
+                                    cpu: 202m
+                                    memory: 512Mi
+                                limits:
+                                    cpu: 2
+                                    memory: 1Gi
+                                livenessProbe:
```
birom, line 45: Is the readinessProbe intentionally missing here?

cinhtau: yes, since we access portal through gateway, we will notice it anyway.

readinessProbe was not important in my eyes, but I was not sure if there is a `management/health` endpoint. If you tell me there is one, I can incoporate the readiness probe into this PR.

birom: thx

cinhtau:
> I would have made 2 commits: 1 for re-factoring and another for the checks, but the results looks.
Agree.

openshift/jobroom2/jobroom-gateway-app/deployment-config.yml
seco-mgabor: There's one for gateway too:
https://alv-ch.atlassian.net/wiki/spaces/JR2/pages/660144164/Management+and+health+endpoints

PR #916, opened 2020-04-28
Df 2196 from master benutzersuche erweitern mit elasticsearch

hamidsoleymani:


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/web/rest/UserInfoResource.java
```diff
@@ -37,6 +46,14 @@ public UserInfoDTO getUserInfo(@RequestParam String eMail) {
         return this.userInfoService.getUserInfo(eMail);
     }
 
+    @GetMapping("/_search")
+    @Timed
+    public ResponseEntity<List<UserInfoDTO>> search(@RequestParam String searchQuery) {
+        Page<UserInfoDTO> userInfoPaged = this.userInfoElasticSearchService.search(searchQuery, PageRequest.of(0, 5));
```
emina-str, line 52: I think we said that search should return 20 results.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/UserInfoElasticSearchService.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private final UserInfoService userInfoService;
+
+    public UserInfoElasticSearchService(UserInfoElasticSearchRepository userInfoElasticSearchRepository,
+                                        UserInfoService userInfoService) {
+        this.userInfoElasticSearchRepository = userInfoElasticSearchRepository;
+        this.userInfoService = userInfoService;
+    }
+
+    public Page<UserInfoDTO> search(String query, Pageable pageable) {
```
emina-str, line 24: I'm not sure does it really work. What about empty string?
Also, the query should check PersonNumber, Email, First and Last Name and Company Name.
For the search method we should right appropriate tests.

hamidsoleymani: 1) Yes it works. If it gets empty string, nothing will be returned.
2) Why should there be a check for those fields?
3) I will think about appropriate test

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/web/rest/UserInfoResource.java
hamidsoleymani: Well, page should always start with 0 and size must be greater than 0. So I took 5.
I can introduce those requestparam fields, so you can set to 0 and 20.

hamidsoleymani:
- method to return one UserInfo (by ID or something) when we go to detail overview; -> it is this method which already exists: @GetMapping("/_search/by-person-nr")

emina-str:
> * method to return one UserInfo (by ID or something) when we go to detail overview; -> it is this method which already exists: @GetMapping("/_search/by-person-nr") and @GetMapping("/{userInfoId}/roles")

Searching by person number can only work for registered stes users, and searching by e-mails can give more then one results, so we need a method to return one valid UserInfo by f.e. id or externalId.

emina-str:
@hamidsoleymani  I also need a method to unregister an user by Id.
Now we have option to unregister by person-number (which functions only for stes) and by e-mail (which can't be ok because of possible multiple UserInfo with same e-mail), so I need a basic `unregister by id` method.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/LoginFormUserDetailsService.java
```diff
@@ -58,28 +58,31 @@ private User getUser(String login) {
...
+            UserInfo newUserInfo = new UserInfo(
                 user.getFirstName(),
                 user.getLastName(),
                 user.getEmail(),
                 user.getLogin(),
                 user.getLangKey()
             );
-            return this.userInfoRepository.saveAndFlush(userInfo1);
+            final UserInfo newUserInfoSaved = this.userInfoRepository.saveAndFlush(newUserInfo);
+            this.applicationEventPublisher.publishEvent(new UserInfoChangedEvent(newUserInfoSaved));
```
emina-str, line 71: I would suggest to name them `UserCreatedEvent` for saving a new object, and `UserUpdatedEvent` for updating existing object. So it matches already existing events for UserInfo -> `UserRegisteredEvent`, `UserUnregisteredEvent`, `UserChangedEvent`.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
```diff
@@ -92,6 +94,7 @@ public void registerAsJobSeeker(LocalDate birthdate, Long personNumber) throws I
         LOGGER.info("Registered user with id: {} as job-seeker", userInfo.getUserExternalId());
         applicationEventPublisher.publishEvent(BusinessLogEvent.of(USER_REGISTERED).withAdditionalData("affectedRole", ROLE_JOBSEEKER.getName()));
         applicationEventPublisher.publishEvent(new UserRegisteredEvent(userInfo, UserInfoRole.JOB_SEEKER));
+        applicationEventPublisher.publishEvent(new UserInfoChangedEvent(userInfo));
```
emina-str, line 97: is there a special reason for sending another event, when there is already registered or unregistered event sent?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/web/rest/UserInfoResource.java
```diff
@@ -26,23 +31,45 @@
...
+    @Deprecated
     @GetMapping("/")
     @Timed
     public UserInfoDTO getUserInfo(@RequestParam String eMail) {
         return this.userInfoService.getUserInfo(eMail);
     }
 
+    @GetMapping("/_search")
+    @Timed
+    public ResponseEntity<List<UserInfoDTO>> search(@RequestParam String searchQuery, @RequestParam(name = "page", defaultValue = "0") int pageNumber, @RequestParam int pageSize) {
```
emina-str, line 52: you can use the same default principle for the page size (in method parameter), and if it's not a big problem, please rename to `query`, `page` and `size`.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoEventListener.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Component
+public class UserInfoEventListener {
+    private final UserInfoElasticSearchService userInfoElasticSearchService;
+
+    public UserInfoEventListener(UserInfoElasticSearchService userInfoElasticSearchService) {
+        this.userInfoElasticSearchService = userInfoElasticSearchService;
+    }
+
+    @EventListener
```
emina-str, line 16: should be Transactional one, I think.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoEventListener.java
```diff
@@ -0,0 +1,20 @@
...
+    void handleUserInfoChangedEvent(UserInfoChangedEvent userChangedEvent) {
```
emina-str, line 17: you can listen the `AbstractUserEvent`, to catch all events for UserInfo.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/UserInfoElasticSearchService.java
```diff
@@ -0,0 +1,42 @@
...
+        QueryStringQueryBuilder queryStringQueryBuilder = queryStringQuery(query);
+
+        return userInfoElasticSearchRepository.search(queryStringQueryBuilder, pageable)
+            .map((userInfoDocument) -> userInfoService.toUserInfoDTO(userInfoDocument.getUserInfo()));
+    }
+
+    public void updateIndex(UserInfo updatedUserInfo) {
+        if (userInfoElasticSearchRepository.existsById(updatedUserInfo.getId().getValue())) {
```
emina-str, line 34: is there a reason to delete it first, and then save? elasticsearch `index()` method is basically for save or update.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoChangedEvent.java
```diff
@@ -0,0 +1,38 @@
+package ch.admin.seco.jobroom.domain;
+
+public class UserInfoChangedEvent extends AbstractUserEvent {
+
+    private final UserInfo userInfo;
```
emina-str, line 5: whole entity object in an Event is not needed.
you already have UserInfoId from AbstractUserEvent, that should be enough.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
hamidsoleymani: Yes. After registered or unregistered events the table user_info is changed, that is why I need to update the elastic index.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/web/rest/UserInfoResource.java
hamidsoleymani: ok.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoEventListener.java
hamidsoleymani: I will do. thanks.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/UserInfoElasticSearchService.java
hamidsoleymani: The documentation says, it is just for save. This is why I did this way.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoChangedEvent.java
hamidsoleymani: I case of save, I need the whole object UserInfo, because i save the whole object. UserInfoId is not enough.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/LoginFormUserDetailsService.java
hamidsoleymani: You are right. But UserRegisteredEvent and UserUnregisteredEvent have different implementation. In UserInfoChangedEvent case, we have no varying implementation. That is why I chose one event for both update and save. I just need an object to encapsulate UserInfo object.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoEventListener.java
hamidsoleymani: I need to listen to UserInfoChangedEvent to get the UserInfo object.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/LoginFormUserDetailsService.java
emina-str: Yes, but we rarely (or never) put whole Entity object in the Event. We usually just call the repository and fetch the object (UserInfo) in the `Listener` part, before saving the elasticsearch document.
Events are for identifying what is happening with our domain objects. It can later on have more implementation then just reindexing the document.
All "UserEvent" extend `AbstractUserEvent` which holds the `UserInfoId` -> so my suggestion is to really create Events to explain what is happening : `UserCreatedEvent`, `UserUpdatedEvent` etc., listen to `AbstractUserEvent` and fetch the UserInfo before saving/indexing UserInfoDocument.

Sending two events which basically represent the same change is not necessary.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
emina-str: Please see above comment about it. I don't think it's needed.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoEventListener.java
emina-str: you should actually fetch the actual UserInfo from db before indexing in elasticsearch, for that only `id` is needed.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/LoginFormUserDetailsService.java
hamidsoleymani: But we have got the UserInfo to save/update. Why should we go to Db and fetch the UserInfo if we already have the object  to save/update? This is  unnecessary loop in my opinion. Because we have done sth. before should not force us to do it exactly the same way.

hamidsoleymani: But we have got the UserInfo to save/update. Why should we go to Db and fetch the UserInfo if we already have the object  to save/update? This is  unnecessary loop in my opinion. Because we have done sth. before should not force us to do it exactly the same way.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoEventListener.java
hamidsoleymani: But I have already the actual UserInfo object to save or update, or do i oversee sth. here?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
```diff
@@ -158,6 +162,8 @@ public RegistrationResultDTO registerAsEmployerOrAgent(String accessCode) throws
 
         result.setSuccess(true);
         userInfo.finishRegistration();
+        applicationEventPublisher.publishEvent(new UserInfoUpdatedEvent(userInfo));
```
emina-str, line 165: I would suggest small change. To add line `userInfo.finishRegistration();` to both switch cases like f.e.
```
            case VALIDATION_EMP:
                result.setEmployerType();
                addCompanyRoleToEiam(userPrincipal);
                addCompanyRoleToSession();
                userInfo.finishRegistration();
                LOGGER.info("Registered user with id: {} as company user", userInfo.getUserExternalId());
               applicationEventPublisher.publishEvent(.....);
                applicationEventPublisher.publishEvent(new UserRegisteredEvent(userInfo, UserInfoRole.COMPANY));
                break;
```
so we can have updated UserInfo before triggering `UserRegisteredEvent`.
Then you can remove the last event : 
`applicationEventPublisher.publishEvent(new UserInfoUpdatedEvent(userInfo));`

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
```diff
@@ -34,9 +47,18 @@
...
 
+    @Autowired
+    ElasticsearchIndexService elasticsearchIndexService;
+
+    @Autowired
+    CompanyRepository companyRepository;
+
     @Before
     public void setUp() {
         this.userInfoRepository.deleteAll();
```
emina-str, line 64: add in setup `this.userInfoElasticSearchRepository.deleteAll();`

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
hamidsoleymani: That was my first idea, but what if we don't enter into switch statement?

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
```diff
@@ -70,21 +96,94 @@ public void testUnregisterByStesNr() throws Exception {
...
         assertThat(savedUserInfo.getEmail()).isEqualTo(eMail);
         assertThat(savedUserInfo.getStesInformation()).isPresent();
         assertThat(savedUserInfo.getStesInformation().get().getPersonNumber()).isEqualTo(personNr);
+        assertThat(savedUserInfo.getRegistrationStatus()).isEqualTo(RegistrationStatus.REGISTERED);
 
         // when-then
         mockMvc.perform(delete("/api/user-info/by-person-nr/?personNr=" + personNr + "&role=" + UserInfoResource.Role.JOB_SEEKER)
             .accept(TestUtil.APPLICATION_JSON_UTF8))
             .andExpect(status().isOk());
+        final Optional<UserInfo> userInfoOptionalFromDb = userInfoRepository.findByEMail("john.doe@test.ch");
```
emina-str, line 110: please use named variable for email and not string.

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
```diff
@@ -49,18 +71,22 @@ public void testUnregisterByEmail() throws Exception {
...
         assertThat(savedUserInfo.getEmail()).isEqualTo(eMail);
         assertThat(savedUserInfo.getStesInformation()).isPresent();
         assertThat(savedUserInfo.getStesInformation().get().getPersonNumber()).isEqualTo(personNr);
+        assertThat(savedUserInfo.getRegistrationStatus()).isEqualTo(RegistrationStatus.REGISTERED);
 
         // when-then
         mockMvc.perform(delete("/api/user-info/?eMail=" + eMail + "&role=" + UserInfoResource.Role.JOB_SEEKER)
             .accept(TestUtil.APPLICATION_JSON_UTF8))
             .andExpect(status().isOk());
+        final Optional<UserInfo> userInfoOptionalFromDb = userInfoRepository.findByEMail("jane.doe@test.ch");
```
emina-str, line 86: please use named variable for email and not string.

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
```diff
@@ -70,21 +96,94 @@ public void testUnregisterByStesNr() throws Exception {
...
+        final Optional<UserInfo> userInfoOptionalFromDb = userInfoRepository.findByEMail("john.doe@test.ch");
+        assertThat(userInfoOptionalFromDb.get().getRegistrationStatus()).isEqualTo(RegistrationStatus.UNREGISTERED);
+        final Optional<UserInfoDocument> userInfoDocumentOptional = userInfoElasticSearchRepository.findById(savedUserInfo.getId().getValue());
+        assertThat(userInfoDocumentOptional.get().getUserInfo().getRegistrationStatus()).isEqualTo(RegistrationStatus.UNREGISTERED);
+    }
+
+    @Test
+    @Transactional
+    public void verifyThatIndexingWorks() {
+        //arrange
```
emina-str, line 119: we have style of `given-when-then` to describe steps in test, please stick to those instead of `arrange-act-assert`.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/RegistrationService.java
hamidsoleymani: ok

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
hamidsoleymani: ok

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
hamidsoleymani: ok

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
hamidsoleymani: ok

jobroom-gateway-app/src/test/java/ch/admin/seco/jobroom/web/rest/UserInfoResourceTest.java
hamidsoleymani: ok

hamidsoleymani:
@birom  ich habe zwei Methoden als @Deprecated annotiert. Sollen Sie noch bleiben oder sollten sie gelöscht werden? Danke

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoId.java
```diff
@@ -1,21 +1,20 @@
...
+import java.util.Objects;
 
 @Embeddable
 @Access(AccessType.FIELD)
 public class UserInfoId implements Serializable {
 
     private final String value;
 
-    UserInfoId() {
+    public UserInfoId() {
```
birom, line 17: Why do change this?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoRepository.java
```diff
@@ -29,4 +31,10 @@
 
     @Query("select u from UserInfo u where u.registrationStatus = 'UNREGISTERED' order by u.id")
     Page<UserInfo> findAllUnregisteredUsers(Pageable pageable);
+
+    @Query("select u from UserInfo u")
+    Stream<UserInfo> streamAll();
```
birom, line 36: We usually use query hints, for the ``streamAll`` methods. Please have a look at the ``FavouriteItemRepository``.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/LoginFormUserDetailsService.java
```diff
@@ -58,28 +58,32 @@ private User getUser(String login) {
...
                 user.getLangKey()
             );
-            return this.userInfoRepository.saveAndFlush(userInfo1);
-        } else {
-            UserInfo userInfo = userInfoOptional.get();
-            boolean hasChanged = userInfo.loginWithUpdate(
+            final UserInfo newUserInfoSaved = this.userInfoRepository.saveAndFlush(newUserInfo);
+            this.applicationEventPublisher.publishEvent(new UserInfoCreatedEvent(newUserInfoSaved));
+            return newUserInfoSaved;
+        } else { //UserInfo is present
```
birom, line 73: Please remove redundant komment.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/saml/DefaultSamlBasedUserDetailsProvider.java
```diff
@@ -90,6 +90,8 @@ private void updateDbUser(EiamEnrichedSamlUser eiamEnrichedSamlUser, UserInfo us
         );
         if (hasChanged) {
             this.applicationEventPublisher.publishEvent(new UserChangedEvent(userInfo));
+        } else {
+            this.applicationEventPublisher.publishEvent(new UserInfoUpdatedEvent(userInfo));
```
birom, line 94: What is the difference between the ``UserChangedEvent`` and ``UserChangedEvent``?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/UserInfoElasticSearchService.java
```diff
@@ -0,0 +1,37 @@
...
+                                        UserInfoService userInfoService) {
+        this.userInfoElasticSearchRepository = userInfoElasticSearchRepository;
+        this.userInfoService = userInfoService;
+    }
+
+    public Page<UserInfoDTO> search(String query, Pageable pageable) {
+        QueryStringQueryBuilder queryStringQueryBuilder = queryStringQuery(query);
+
+        return userInfoElasticSearchRepository.search(queryStringQueryBuilder, pageable)
+            .map((userInfoDocument) -> userInfoService.toUserInfoDTO(userInfoDocument.getUserInfo()));
```
birom, line 30: The toUserInfoDTO method is static. We do not need the UserInfoService instance here.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/saml/DefaultSamlBasedUserDetailsProvider.java
emina-str: @birom , I can maybe explain this.
`UserChangedEvent` is an old event we created, it is for the changes in the fields : firstName, lastName, eMail . This Event is consumed also in Onlineformservice, where we have to change then the name of the User.
`UserUpdatedEvent` is a new event now, a general update event, when whatever fields get updated (consumed only in Elasticsearch).

Maybe it would be clearer if we rename the `UserChangedEvent` to something like `UserPersonalDataUpdatedEvent` or something like that?

hamidsoleymani: In case any field of UserInfo is changed, "modifiedAt" field gets updated. We have to update as well elastic index. modifiedAt is shown in FE, and it must be the updated as well in in FE

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfo.java
```diff
@@ -122,7 +120,7 @@ public UserInfo(String firstName, String lastName, String email, String userExte
         this.touch();
     }
 
-    private UserInfo() {
+    public UserInfo() {
```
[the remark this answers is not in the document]
hamidsoleymani: ok

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoId.java
hamidsoleymani: ok

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/security/LoginFormUserDetailsService.java
hamidsoleymani: ok

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/UserInfoElasticSearchService.java
hamidsoleymani: ok

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/UserInfoRepository.java
hamidsoleymani: In jobbroom ``CompanyRepository``, ``OrganizationRepository`` and  ``UserRepository`` don't use QueryHints. That is why I did it that way. What shall we do?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/userinfo/UserInfoRepository.java
```diff
@@ -29,4 +34,7 @@
 
     @Query("select u from UserInfo u where u.registrationStatus = 'UNREGISTERED' order by u.id")
     Page<UserInfo> findAllUnregisteredUsers(Pageable pageable);
+
+    @Query("select u from UserInfo u")
+    Stream<UserInfo> streamAll();
```
birom, line 39: This causes a "n +1 query problem" if we do not use an ``EntityGraph``.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoDocument.java
```diff
@@ -0,0 +1,33 @@
...
+
+
+import ch.admin.seco.jobroom.domain.userinfo.UserInfo;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.elasticsearch.annotations.Document;
+
+import java.io.Serializable;
+
+
+@Document(indexName = "user-info", type = "user-info", shards = 1, replicas = 0)
```
birom, line 11: Why do not we have a json mapping file?

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/UserInfoService.java
```diff
@@ -115,7 +123,7 @@ public void sendAllRegisteredStes() {
         do {
             pageResult = this.userInfoRepository.findAllRegisteredUsers(nextPageable);
             pageResult.filter(userInfo -> userInfo.getStesInformation().isPresent())
-                .forEach(userInfo -> applicationEventPublisher.publishEvent(new UserRegisteredEvent(userInfo, UserInfoRole.JOB_SEEKER)));
+                .forEach(userInfo -> applicationEventPublisher.publishEvent(new UserInfoRegisteredEvent(userInfo, UserInfoRole.JOB_SEEKER)));
```
emina-str, line 126: couldn't we use `publish(new UserInfoRegisteredEvent(userInfo, UserInfoRole.JOB_SEEKER))` ?
also, under with `UserInfoUnregisteredEvent` .. so we can remove `applicationEventPublisher` in constructor.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/domain/userinfo/UserInfoRepository.java
hamidsoleymani: fixed

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/search/UserInfoDocument.java
hamidsoleymani: added.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/UserInfoService.java
hamidsoleymani: We could introduce two methods in UserInfo:
1)  `public void publishWithRole(UserInfoRole userInfoRole){
        publish(new UserInfoRegisteredEvent(this, userInfoRole));
    }`
2) ` public void publishWithoutRole(){
        publish(new UserInfoUnregisteredEvent(this));
    }`
Then we could use in UserInfoService: 1) `.forEach(userInfo -> userInfo.publishWithRole(UserInfoRole.JOB_SEEKER));`
2) `pageResult.forEach(userInfo -> userInfo.publishWithoutRole());`
and we can get rid of `applicationEventPublisher`

What do you think @birom  and @emina-str

emina-str: I'm not sure about that, I think `publish()` shouldn't care about the event and parameters we send through.

hamidsoleymani: Hmm, I am not sure which publish method you are talking about. I am talking about DomainEventPublisher.publish and this publish doesn't care about the event.

hamidsoleymani: I see, we don't need those intermediate methods in UserInfo, which I proposed. We can directly call DomainEventPublisher.publish...

PR #919, opened 2020-05-29
DF-2307 : Unregister User on CandidateDeletedEvent

emina-str:
- fix the problem with NonUniqueResult
- add more information for the e-mail that is sent


jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/MailService.java
```diff
@@ -97,11 +97,12 @@ public void sendAccessCodeLetterMail(String emailAddress, UserInfo userInfo) {
         sendEmail(emailAddress, subject, content, true, true, attachmentFilename, pathToPdf);
     }
 
-    public void sendStesUnregisteringMail(String stesEmail, String recipient) {
-        log.info("Send an email for unregister a candidate with email {} to {} ", stesEmail, recipient);
-        Locale locale = Locale.forLanguageTag("de");
+    public void sendStesUnregisteringMail(UserInfo userInfo, String recipient) {
+        log.info("Send an email for unregistering a candidate with email {} to {} ", userInfo.getEmail(), recipient);
+        Locale locale = LocaleContextHolder.getLocale();
         Context context = new Context(locale);
```
hamidsoleymani, line 103: Here is locale always "de".  Should `LocaleContextHolder.getLocale()`  also return always "de"?

emina-str: > getLocale() : Return the Locale associated with the current thread, if any, or the system default Locale otherwise.

I think it is a better approach then adding a `de` language. Also, f.e. on my machine e-mail tests for content were failing, even though on travis they were successful because of this system default Locale. Those e-mails go to service desk, and will have the system default language.
On other point, the unregistering e-mail only have English translations for all languages.

emina-str: I reverted for the method which is not subject of this PR `sendAccessCodeLetterMail()`, but for this `sendStesUnregisteringMail()` I would keep it.

jobroom-gateway-app/src/main/java/ch/admin/seco/jobroom/service/MailService.java
```diff
@@ -97,11 +97,12 @@ public void sendAccessCodeLetterMail(String emailAddress, UserInfo userInfo) {
...
-        context.setVariable("candidateEmail", stesEmail);
+        context.setVariable("candidateEmail", userInfo.getEmail());
+        context.setVariable("candidatePersonNumber", userInfo.getStesInformation().get().getPersonNumber());
```
birom, line 105: Please assert/verify at the beginning of the method call that  ``userInfo.getStesInformation()`` has a value.

emina-str: added.

